PR #1281, opened 2022-05-19
Introduce paging to CRUD.

BearLemma:
Brings paging to CRUD.
I think that some things still remain to be done, namely:

- [ ] docusaurus docs
- [ ] `previous_page` in response and `page_before=cursor` parameter
- [x] lit test complementing the paging unit tests.

I guess that 1) should be done in this PR, but I would do 2 and 3 in later PRs to minimize the size of this PR, it's pretty big anyway.

### DIFF
--- a/cli/tests/lit/crud.node
+++ b/cli/tests/lit/crud.node
@@ -198,9 +198,9 @@ $CURL -X PUT -d '{"age":98422}' $CHISELD_HOST/dev/persons/cef5d492-d7e3-4c45-9a5
 $CURL -X PUT -d '{"age":2938}' $CHISELD_HOST/dev/persons/cef5d492-d7e3-4c45-9a55-5929b9ab8292 # Repeated PUT of the same id.
 # CHECK: HTTP/1.1 200 OK
 # CHECK: "age": 2938
-echo there is just `$CURL --no-include $CHISELD_HOST/dev/persons | jq 'length'` entry
+echo there is just `$CURL --no-include $CHISELD_HOST/dev/persons | jq '.results | length'` entry
 # CHECK: there is just 1 entry
-$CURL --no-include $CHISELD_HOST/dev/persons | jq '.[] | .age'
+$CURL --no-include $CHISELD_HOST/dev/persons | jq '.results | .[] | .age'
 # CHECK: 2938
 
 $CURL -X PUT -d '{"age":98422}' $CHISELD_HOST/dev/persons/ # PUT without ID.
--- a/cli/tests/lit/crud.node
+++ b/cli/tests/lit/crud.node
@@ -61,7 +61,7 @@ $CURL "$CHISELD_HOST/dev/persons?sort=-age"
 count_results () {
     results=$(cat)
     printf '%s' "$results" \
-        | grep -o "last_name" \
+        | grep -o "\"last_name\":" \
         | wc -l \
         | xargs printf 'Number of names: %d\n';
     printf '%s' "$results";
@@ -148,6 +148,26 @@ $CURL $CHISELD_HOST/dev/persons/$id # GET by id
 # CHECK: "first_name": "Dawn"
 $CURL $CHISELD_HOST/dev/persons/i-will-never-exist-ya-know # GET by invalid id
 # CHECK: HTTP/1.1 404 Not Found
+
+## Test paging
+
+resp=$($CURL --no-include "$CHISELD_HOST/dev/persons?sort=last_name&page_size=2&.last_name~ne=Brown")
+echo "$resp" | count_results
+# CHECK: Number of names: 2
+# CHECK: "last_name": "Anderson"
+# CHECK: "last_name": "Chong"
+
+resp=$($CURL --no-include $(echo "$resp" | jq -r '.next_page'))
+echo "$resp" | count_results
+# CHECK: Number of names: 2
+# CHECK: "last_name": "Draper"
+# CHECK: "last_name": "Elgin"
+
+resp=$($CURL --no-include $(echo "$resp" | jq -r '.next_page'))
+echo "$resp" | count_results
+# CHECK: Number of names: 0
+
+## Test delete
 
 $CURL -X DELETE $CHISELD_HOST/dev/persons/$id # DELETE by id
 # CHECK: HTTP/1.1 200 OK
--- a/cli/tests/lit/pol-anon.deno
+++ b/cli/tests/lit/pol-anon.deno
@@ -145,7 +145,7 @@ $CURL --no-include -d '{
 
 # CHECK: Chiselstrike
 
-$CURL --no-include $CHISELD_HOST/dev/companies | jq '.[] | .name, .ceo.firstName, .ceo.lastName, .accountant, .secretSauce'
+$CURL --no-include $CHISELD_HOST/dev/companies | jq '.results | .[] | .name, .ceo.firstName, .ceo.lastName, .accountant, .secretSauce'
 
 # CHECK: Chiselstrike
 # CHECK: xxxxx
--- a/server/src/datastore/crud.rs
+++ b/server/src/datastore/crud.rs
@@ -7,6 +7,7 @@ use anyhow::{Context, Result};
 use deno_core::futures;
 use futures::{Future, StreamExt};
 use serde_derive::{Deserialize, Serialize};
+use serde_json::json;
 use std::sync::Arc;
 use url::Url;
 
@@ -23,7 +24,7 @@ pub(crate) fn run_query(
     params: QueryParams,
     query_engine: Arc<QueryEngine>,
     tr: TransactionStatic,
-) -> impl Future<Output = Result<Vec<JsonObject>>> {
+) -> impl Future<Output = Result<JsonObject>> {
     let fut = run_query_impl(context, params, query_engine, tr);
     async move { fut?.await }
 }
@@ -33,7 +34,7 @@ fn run_query_impl(
     params: QueryParams,
     query_engine: Arc<QueryEngine>,
     tr: TransactionStatic,
-) -> Result<impl Future<Output = Result<Vec<JsonObject>>>> {
+) -> Result<impl Future<Output = Result<JsonObject>>> {
     let base_type = &context
         .ts
         .lookup_object_type(&params.type_name, &context.api_version)
@@ -45,13 +46,54 @@ fn run_query_impl(
     let stream = query_engine.query(tr.clone(), query_plan)?;
 
     Ok(async move {
-        stream
+        let results = stream
             .collect::<Vec<_>>()
             .await
             .into_iter()
             .collect::<Result<Vec<_>>>()
-            .context("failed to collect result rows from the database")
+            .context("failed to collect result rows from the database")?;
+
+        let mut ret = JsonObject::new();
+        if !results.is_empty() && query.page_size == results.len() as u64 {
+            let last_element = results.last().unwrap();
+            let next_page = next_page_url(&params.url, &query, last_element)?;
+            ret.insert("next_page".into(), json!(next_page));
+        }
+
+        ret.insert("results".into(), json!(results));
+        Ok(ret)
     })
+}
+
+fn next_page_url(url: &Url, query: &Query, last_element: &JsonObject) -> Result<Url> {
+    assert!(!query.sort.keys.is_empty());
+
+    let mut axes = vec![];
+    for key in query.sort.keys.iter().cloned() {
+        let value = last_element
+            .get(&key.field_name)
+            .cloned()
+            .with_context(|| {
+                format!("failed to create cursor axis for field `{}", key.field_name)
+            })?;
+        axes.push(CursorAxis { key, value });
+    }
+    let cursor = serde_json::to_vec(&axes).context("failed to serialize cursor to json")?;
+    let cursor = base64::encode(cursor);
+
+    let mut next_url = url.clone();
+    next_url.set_query(Some(""));
+    for (key, value) in url.query_pairs() {
+        if key == "page_after" || key == "sort" {
+            continue;
+        }
+        next_url.query_pairs_mut().append_pair(&key, &value);
+    }
+    next_url
+        .query_pairs_mut()
+        .append_pair("page_after", &cursor);
+
+    Ok(next_url)
 }
 
 /// Constructs Delete Mutation from CRUD url.
@@ -77,8 +119,9 @@ pub(crate) fn delete_from_url(c: &RequestContext, type_name: &str, url: &str) ->
 
 /// Query is used in the process of parsing crud url query to rust representation.
 struct Query {
-    limit: Option<u64>,
+    page_size: u64,
     offset: Option<u64>,
+    cursor: Option<Cursor>,
     sort: SortBy,
     /// Filters restricting the result set. They will be joined in AND-fashion.
     filters: Vec<Expr>,
@@ -87,8 +130,9 @@ struct Query {
 impl Query {
     fn new() -> Self {
         Query {
-            limit: None,
+            page_size: 1000,
             offset: None,
+            cursor: None,
             sort: SortBy {
                 keys: vec![SortKey {
                     field_name: "id".into(),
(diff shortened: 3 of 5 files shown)


cli/tests/lit/pol-anon.deno
```diff
@@ -136,9 +136,13 @@ export default crud(
   "",
   {
     defaultCreateResponse: (body: unknown, status: number) => {
+      let results = (body as unknown as Record<string, unknown>)["results"];
```
dekimir, line 139: Is `as unknown` really required here?

server/src/datastore/crud.rs
```diff
@@ -44,15 +45,57 @@ fn run_query_impl(
...
+            })?;
+        axes.push(CursorAxis { key, value });
+    }
+    let cursor = serde_json::to_vec(&axes).context("failed to serialize cursor to json")?;
+    let cursor = base64::encode(cursor);
+
+    let mut next_url = url.clone();
+    next_url.set_query(Some(""));
+    for (key, value) in url.query_pairs() {
+        if key == "page_after" || key == "page_before" || key == "sort" {
```
dekimir, line 87: Why skip `sort`?

server/src/datastore/crud.rs
```diff
@@ -44,15 +45,57 @@ fn run_query_impl(
...
+
+        let mut next_page = None;
+        if !results.is_empty() && query.page_size == results.len() as u64 {
+            let last_element = results.last().unwrap();
+            next_page = Some(next_page_url(&params.url, &query, last_element)?)
+        }
+
+        let mut ret = JsonObject::new();
+        ret.insert("results".into(), json!(results));
+        ret.insert("next_page".into(), json!(next_page));
```
dekimir, line 63: No `prev_page`?

server/src/datastore/crud.rs
```diff
@@ -98,19 +148,25 @@ impl Query {
...
                 "offset" => {
                     let o = value.parse().with_context(|| {
                         format!("failed to parse offset. Expected u64, got '{}'", value)
                     })?;
                     q.offset = Some(o);
                 }
+                "page_after" => {
+                    anyhow::ensure!(
+                        q.cursor.is_none(),
+                        "only one occurrence of page_before/page_after is allowed."
```
dekimir, line 166: Are there any tests with `page_before`?  Is it even handled when provided?

server/src/datastore/crud.rs
```diff
@@ -520,6 +675,52 @@ mod tests {
...
+                    assert!(r["next_page"].is_null());
+                } else {
+                    let next_page = r["next_page"].as_str().unwrap();
+                    page_url = Url::parse(next_page).unwrap();
+                }
+            }
+            assert_eq!(all_names, vec!["Alan", "Alex", "John", "Steve"]);
+        }
+        {
+            let r = run_query("Person", url("sort=name&page_size=5"), qe)
```
dekimir, line 716: We really need more testcases than `page_size=1` and `page_size=<big number>`.  Most of the real usecases will fall in between the two.

server/src/datastore/crud.rs
```diff
@@ -520,6 +675,52 @@ mod tests {
...
             }
         }
+
+        // Test cursors
+        {
+            let mut page_url = url("page_size=1");
+            let mut all_names = vec![];
+            for i in 0..5 {
+                let r = run_query("Person", page_url.clone(), qe).await.unwrap();
+                let names = collect_names(&r);
```
dekimir, line 685: Shouldn't we assert r's size here?

server/src/datastore/crud.rs
BearLemma: Because when we are generating url for next page, the sort cannot be changed and is included in the cursor.

server/src/datastore/crud.rs
BearLemma: See description of the PR, I would prefer to implement prev_page in the next PR.

server/src/datastore/crud.rs
BearLemma: Same as above, deferred for another PR.

server/src/datastore/crud.rs
BearLemma: Yup, added assert to the branch where the names are non-empty.

server/src/datastore/crud.rs
BearLemma: Good point, added page_size = 2 to the second test case.

cli/tests/lit/pol-anon.deno
BearLemma: Good point, fixed.

server/src/datastore/crud.rs
```diff
@@ -44,15 +45,57 @@ fn run_query_impl(
...
-        stream
+        let results = stream
             .collect::<Vec<_>>()
             .await
             .into_iter()
             .collect::<Result<Vec<_>>>()
-            .context("failed to collect result rows from the database")
+            .context("failed to collect result rows from the database")?;
+
+        let mut next_page = None;
```
espindola, line 55: I would replace `mut` with `let next_page = if ... {} else {None};`

Also, are you sure the right comparison to use is '=='? It should be '<', no?

espindola (commented):
This is missing a test where we produce more than one page.

server/src/datastore/crud.rs
BearLemma: Ah, good point. Fixed.
Yeah, I'm pretty sure about it. Only if you fulfill the page_size/limit you can expect to have a next page.

BearLemma:
> This is missing a test where we produce more than one page.

Isn't that covered by those tests: https://github.com/chiselstrike/chiselstrike/pull/1281/files#diff-20558519d1818bea6999042f85ad4ee0babd4ee084e3964bbeee7752021b0ee1R683 ?

espindola:
> > This is missing a test where we produce more than one page.
> 
> Isn't that covered by those tests: https://github.com/chiselstrike/chiselstrike/pull/1281/files#diff-20558519d1818bea6999042f85ad4ee0babd4ee084e3964bbeee7752021b0ee1R683 ?

Ah, could you add a .deno test too? This is a high level feature, so it is probably good to have an end to end test.

server/src/datastore/crud.rs
dekimir: Ok.

server/src/datastore/crud.rs
dekimir: Then why is this PR referencing it?  I'm primarily thinking of https://github.com/chiselstrike/chiselstrike/pull/1281/files#diff-20558519d1818bea6999042f85ad4ee0babd4ee084e3964bbeee7752021b0ee1R88

BearLemma: I can omit it, but I put it in place so that I don't forget it when I add page_before in the next PR.

dekimir: Can you track such to-do items in some other way?  I prefer a clean, logical PR with simple-to-understand code.  Random pollution like this makes it harder to review, harder to explain to newcomers, harder to backport, and harder to tell if a feature is implemented at this commit or not.

server/src/datastore/crud.rs
dekimir: OK

cli/tests/integration_tests.rs
```diff
@@ -159,7 +159,7 @@ fn run_tests(opt: Opt, optimize: bool) -> bool {
                     let i = rayon::current_thread_index().unwrap_or(0) + 1;
                     config.test_paths = vec![test_path.clone()];
                     config.truncate_output_context_to_number_of_lines = Some(500);
-                    config.always_show_stdout = false;
+                    config.always_show_stdout = true;
```
dekimir, line 162: Is this deliberate?  I thought it was for debugging after reading this: https://github.com/chiselstrike/chiselstrike/pull/1327#discussion_r882653065

I don't see how the new tests you added would depend on this being `true`.

server/src/datastore/crud.rs
BearLemma: Yup, that makes sense, fixed.

cli/tests/integration_tests.rs
BearLemma: Ah, this was accidental, fixed.

dekimir (approved):
LGTM, though it looks like it wasn't rebased on latest `main` (see, eg, [this](https://github.com/chiselstrike/chiselstrike/pull/1281/files#diff-e0bca91fbd8a766ba1468f3e8fb4ad2976ecaf383bde31f572741c907ff725dd)).

Also, failed CI build. :)

BearLemma:
> LGTM, though it looks like it wasn't rebased on latest `main` (see, eg, [this](https://github.com/chiselstrike/chiselstrike/pull/1281/files#diff-e0bca91fbd8a766ba1468f3e8fb4ad2976ecaf383bde31f572741c907ff725dd)).
> 
> Also, failed CI build. :)

So the CI fails because of the lit test that actually uses the `next_page` link, but fails because of the host problem (the `next_url` in our tests points to `chiselstrike.com`). So we can either disable the test for the time being, merge as is and follow up with solution to our host problem, or the other way around. I don't really have a preference in this case :man_shrugging:

espindola:
> > LGTM, though it looks like it wasn't rebased on latest `main` (see, eg, [this](https://github.com/chiselstrike/chiselstrike/pull/1281/files#diff-e0bca91fbd8a766ba1468f3e8fb4ad2976ecaf383bde31f572741c907ff725dd)).
> > Also, failed CI build. :)
> 
> So the CI fails because of the lit test that actually uses the `next_page` link, but fails because of the host problem (the `next_url` in our tests points to `chiselstrike.com`). So we can either disable the test for the time being, merge as is and follow up with solution to our host problem, or the other way around. I don't really have a preference in this case man_shrugging

I would prefer to fix our host problem before or in this PR. Either configuring the host or using a relative URL, depending on what we gather from web developer preferences.

BearLemma:
> > > LGTM, though it looks like it wasn't rebased on latest `main` (see, eg, [this](https://github.com/chiselstrike/chiselstrike/pull/1281/files#diff-e0bca91fbd8a766ba1468f3e8fb4ad2976ecaf383bde31f572741c907ff725dd)).
> > > Also, failed CI build. :)
> > 
> > 
> > So the CI fails because of the lit test that actually uses the `next_page` link, but fails because of the host problem (the `next_url` in our tests points to `chiselstrike.com`). So we can either disable the test for the time being, merge as is and follow up with solution to our host problem, or the other way around. I don't really have a preference in this case man_shrugging
> 
> I would prefer to fix our host problem before or in this PR. Either configuring the host or using a relative URL, depending on what we gather from web developer preferences.

Fixed :relaxed:  Tests should be passing now.

cli/tests/lit/crud.node
```diff
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
 
-# RUN: sh -e @file
+# RUN: bash -e @file
```
espindola, line 3: Why do we need bash? Not that bash in OS X is really old.

cli/tests/lit/crud.node
```diff
@@ -149,6 +149,26 @@ $CURL $CHISELD_HOST/dev/persons/$id # GET by id
...
 
+## Test paging
+
+resp=$($CURL --no-include "$CHISELD_HOST/dev/persons?sort=last_name&page_size=2&.last_name~ne=Brown")
+echo "$resp" | count_results
+# CHECK: Number of names: 2
+# CHECK: "last_name": "Anderson"
+# CHECK: "last_name": "Chong"
+
+resp=$($CURL --no-include `echo "$resp" | jq -r '.next_page'`)
```
espindola, line 160: `$()` is easier to read than using backquotes

cli/tests/lit/crud.node
```diff
@@ -149,6 +149,26 @@ $CURL $CHISELD_HOST/dev/persons/$id # GET by id
...
+echo "$resp" | count_results
+# CHECK: Number of names: 2
+# CHECK: "last_name": "Draper"
+# CHECK: "last_name": "Elgin"
+
+resp=$($CURL --no-include `echo "$resp" | jq -r '.next_page'`)
+echo "$resp" | count_results
+# CHECK: Number of names: 0
```
espindola, line 168: Is this expected? There is always a last page that is empty?

BearLemma: Only when `results_count % page_size == 0`. To avoid this problem, I was thinking of always retrieving `page_size + 1` elements, but I don't want to do it in this PR.

cli/tests/lit/crud.node
BearLemma: Ah, good catch.

cli/tests/lit/crud.node
BearLemma: Ok

server/src/datastore/crud.rs
```diff
@@ -44,15 +45,58 @@ fn run_query_impl(
...
-        stream
+        let results = stream
             .collect::<Vec<_>>()
             .await
             .into_iter()
             .collect::<Result<Vec<_>>>()
-            .context("failed to collect result rows from the database")
+            .context("failed to collect result rows from the database")?;
+
+        let next_page = if !results.is_empty() && query.page_size == results.len() as u64 {
```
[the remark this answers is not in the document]
BearLemma: Yup, that is true, but I guess I prefer the original version as it feels a little more explicit to me (it says that we generate next_page url only if there are at least some results).

server/src/datastore/crud.rs
```diff
@@ -76,19 +126,26 @@ pub(crate) fn delete_from_url(c: &RequestContext, type_name: &str, url: &str) ->
 
 /// Query is used in the process of parsing crud url query to rust representation.
 struct Query {
-    limit: Option<u64>,
+    page_size: u64,
     offset: Option<u64>,
```
espindola, line 130: We drop limit but not offset. When do we use offset instead of Cursor?

server/src/datastore/crud.rs
```diff
@@ -76,19 +126,26 @@ pub(crate) fn delete_from_url(c: &RequestContext, type_name: &str, url: &str) ->
...
-    sort: Option<SortBy>,
+    cursor: Option<Cursor>,
+    sort: SortBy,
```
espindola, line 132: We now always sort, even when not paging. Wouldn't it get too expensive?

server/src/datastore/crud.rs
```diff
@@ -44,15 +46,63 @@ fn run_query_impl(
...
+        let next_page = if !results.is_empty() && query.page_size == results.len() as u64 {
+            let last_element = results.last().unwrap();
+            Some(next_page_url(&params.url, &host, &query, last_element)?)
+        } else {
+            None
+        };
+
+        let mut ret = JsonObject::new();
+        ret.insert("results".into(), json!(results));
+        ret.insert("next_page".into(), json!(next_page));
```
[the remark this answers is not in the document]
BearLemma: Good point, fixed.

server/src/datastore/crud.rs
BearLemma: I've dropped `limit` only internally, it's still preserved in the API. I wanted to provide some grace period for FE dogfood guys to transition so that we don't get similar issues as when I changed booleans to use boolean type instead of int..

server/src/datastore/crud.rs
BearLemma: Well we are always paging now. Plus if no sort is specified, It should use the ID sort which should always have an index. Perhaps there should be a way to bypass this? But it feels like a product decision so cc @dekimir @penberg

server/src/datastore/crud.rs
```diff
@@ -123,30 +184,123 @@ impl Query {
...
-            ops.push(QueryOp::Take { count: limit });
-        }
+        ops.push(QueryOp::Take {
+            count: self.page_size,
+        });
         Ok(ops)
     }
 }
 
+fn ensure_sort_by_id(mut sort: SortBy) -> SortBy {
```
espindola, line 213: Why not take a `&mut SortBy`?

espindola: I think `&mut` is the canonical way to do it in rust. Since the borrow checker guarantees it is safe.

BearLemma: Fixed

espindola:
This needs a rebase.

server/src/datastore/crud.rs
dekimir: See https://github.com/chiselstrike/chiselstrike/pull/1345#issuecomment-1143010496

PR #1282, opened 2022-05-19
Move CRUD query processing entirely to crud.rs.

BearLemma:


server/src/deno.rs
```diff
@@ -667,46 +666,38 @@ fn op_chisel_get_secret(op_state: &mut OpState, key: String) -> Result<Option<se
...
-            &params.url,
-        )?;
-
-        let transaction = current_transaction(op_state);
-        let query_engine = query_engine_arc(op_state);
-        query_engine.query(transaction, query_plan)?
+        (
+            current_transaction(op_state),
+            query_engine_arc(op_state),
+            // FIXME: Cloning here is inefficient but otherwise we have state reference
```
espindola, line 681: What you need to do is clarify the lifetime in the implementation. Change it to:

```
pub(crate) fn run_query(
    context: &RequestContext<'_>,
    params: QueryParams,
    query_engine: Arc<QueryEngine>,
    tr: TransactionStatic,
) -> impl Future<Output = Result<Vec<JsonObject>>> {
    let stream = make_stream(context, params, query_engine, tr);
    async {
        stream?
            .collect::<Vec<Result<JsonObject>>>()
            .await
            .into_iter()
            .collect::<Result<_>>()
            .context("failed to collect result rows from the database")
    }
}

fn make_stream(
    context: &RequestContext<'_>,
    params: QueryParams,
    query_engine: Arc<QueryEngine>,
    tr: TransactionStatic,
) -> Result<impl Stream<Item = Result<JsonObject>>> {
    let url = Url::parse(&params.url)
        .with_context(|| format!("crud endpoint failed to parse url: '{}'", params.url))?;
    let query = Query::from_url(context, &params.type_name, &url)?;
    let stream = {
        let query_plan = query.make_query_plan(context, &params.type_name)?;
        query_engine.query(tr.clone(), query_plan)?
    };
    Ok(stream)
}
```

espindola: With that the code in deno.rs becomes:

```
#[op]
async fn op_chisel_crud_query(
    state: Rc<RefCell<OpState>>,
    params: crud::QueryParams,
    context: ChiselRequestContext,
) -> Result<Vec<JsonObject>> {
    // Contextualize stream creation to prevent state RC borrow living across await                                                                                                                                             
    let fut = {
        let op_state = &state.borrow();
        let transaction = current_transaction(op_state);
        let query_engine = query_engine_arc(op_state);

        crud::run_query(
            &RequestContext {
                policies: current_policies(op_state),
                ts: current_type_system(op_state),
                api_version: context.api_version,
                user_id: context.user_id,
                path: context.path,
            },
            params,
            query_engine,
            transaction,
        )
    };
    fut.await
}
```

espindola:
Even with the suggested changes above it still looks like this PR does quite a bit more than "Move CRUD query processing". A lot of implementation function are changed for example. Could you make this PR really just move the processing? That is, the change to deno.rs should be as above, but the change to crud.rs should be minimized.

server/src/deno.rs
BearLemma: Fixed. Thx!

BearLemma:
> Even with the suggested changes above it still looks like this PR does quite a bit more than "Move CRUD query processing". A lot of implementation function are changed for example. Could you make this PR really just move the processing? That is, the change to deno.rs should be as above, but the change to crud.rs should be minimized.

Fair enough, created https://github.com/chiselstrike/chiselstrike/pull/1292

espindola:
Please describe what Query is and why it is needed. It looks like it is a more strict version of Vec<QueryOp>, but what is the idea, to replace Vec<QueryOp> completely? Also, this PR can still be reduced by not changing auxiliary functions as much and by splitting: See https://github.com/chiselstrike/chiselstrike/pull/1296 for example.

BearLemma:
> Please describe what Query is and why it is needed. It looks like it is a more strict version of Vec, but what is the idea, to replace Vec completely? Also, this PR can still be reduced by not changing auxiliary functions as much and by splitting: See #1296 for example.

Right. The reason for a strict `Query` is that when I implement paging, I need to play with sort/page_size and it's very inconvenient and opaque to search for it in a vector of ops. This way, it's completely explicit what is the structure of ops that will end up in the ops vector.

Fair point about the auxiliary functions, will look into it.

espindola:
> Right. The reason for a strict `Query` is that when I implement paging, I need to play with sort/page_size and it's very inconvenient and opaque to search for it in a vector of ops. This way, it's completely explicit what is the structure of ops that will end up in the ops vector.

That would be a good PR description :-)

athoscouto (commented):
:pick:: a couple of related suggestions only because of a tiny nitpick. Feel free to ignore it.

The main driver was that I found it weird that `make_query_plan` required `RequestContext`.
`make_query_plan` is the only thing that this class does and `RequestContext` is already provided on `from_url`.

One could also argue that `make_query_plan` is doing two things when should be doing one. Or maybe that is easier to unit test this way.

Anyway, nothing major or mandatory.

The code looks good!

BearLemma:
> pick: a couple of related suggestions only because of a tiny nitpick. Feel free to ignore it.
> 
> The main driver was that I found it weird that `make_query_plan` required `RequestContext`. `make_query_plan` is the only thing that this class does and `RequestContext` is already provided on `from_url`.
> 
> One could also argue that `make_query_plan` is doing two things when should be doing one. Or maybe that is easier to unit test this way.
> 
> Anyway, nothing major or mandatory.
> 
> The code looks good!

I've put the context reference inside the class. The main reason I didn't do it in the first place was the additional lifetime complexity as things in RequestContext are of referential nature and hence its lifetime must be guarded. Not sure if this version is better or not. @espindola what do you think?

espindola (approved):

PR #1286, opened 2022-05-19
Fix unnecessary rebuilds

espindola:
If a build.rs ever prints a cargo:rerun-if-changed= pointing to a non
existing file, cargo will rerun that build.rs.

This library is used from other directories, so the relative paths are
wrong, causing tsc_compile to always be rebuilt.

Since this code is using include_str!, the rerun-if-changed was at
most redundant anyway, so drop it.


BearLemma:
Tried it, tsc compilation seems to be working, lgtm.

BearLemma (approved):

PR #1290, opened 2022-05-20
Auto indexing support

penberg:
This pull request implements auto-indexing based on predicates.

The `chiselc` compiler now supports a `index-json` target, which outputs
predicates. For example, the following TypeScript code:

```typescript
 Person.cursor().filter({ name: name, age: age });
 ```

outputs the following predicate indexes:

 ```json
 [{
        "entity_type": "Person",
        "properties": ["name", "age"]
 }]
 ```

The `chisel apply` command invokes `chiselc` to discover predicates from
all the endpoints, and passes that information over `chiseld`, which
treats each predicate as an index. This is a crude approach that
generates more indexes that might be needed.

We need to imprive this in the future by, for example, requiring an
endpoint to have some specific SLA to have its predicates indexed or by
providing dynamic runtime feedback from the database for `chiseld` to
determine which candidate indexes should be indexed.

### CODE REVIEW

**Comment by penberg** [comment]
As discussed with @dekimir, let's put this behind a feature flag that is disabled by default

**Comment by penberg** [comment]
@dekimir There's a feature flag now.

File: chiselc/docs/indexing.md [hunk 1]
```diff
@@ -0,0 +1,23 @@
...
+
+```json
+[{
+	"entity_type": "Person",
+	"properties": ["name", "age"]
+}]
+```
+
+The `chiselc` compiler makes no guesses on the access patterns of the predicate.
+It is up to a runtime or a query profiler to decide wheter to create the index or not.
```
dekimir, line 23: I guess this answers my question above: it's a list of all properties referenced in filters.  Then the name `index-json` is not the best; a name like `filtered-properties.json` would describe the file's purpose better.

server/src/datastore/engine.rs
```diff
@@ -317,6 +324,54 @@ impl QueryEngine {
...
+        indexes: &[DbIndex],
+    ) -> Result<()> {
+        for index in indexes {
+            let idx_name = index
+                .name()
+                .context("index must have a name at a time of table creation")?;
+            let columns = index.fields.iter().join(", ");
+            let create_index = format!(
+                r#"
+                CREATE INDEX IF NOT EXISTS "{idx_name}" ON "{}" ({columns});
```
dekimir, line 345: So we create the same index for `e => e.col1==e.col2` and `e => e.col1>2 && e.col2<5`?  That seems dubious.

BearLemma: I don't get how it's related to this line of code? And from what do you deduce this conclusion? I'm not aware of anything that would imply this behavior :thinking:

dekimir: I tracked how the rest of the code constructs the value of `({columns})`.

chiselc/src/parse.rs
```diff
@@ -83,32 +83,37 @@ pub fn compile(
...
-        Target::JavaScript => {
-            let globals = Globals::default();
-            GLOBALS.set(&globals, || {
-                let top_level_mark = Mark::fresh(Mark::root());
-                module.fold_with(&mut swc_ecmascript::transforms::typescript::strip(
-                    top_level_mark,
-                ))
-            })
+    match target {
+        Target::JavaScript | Target::TypeScript => {
```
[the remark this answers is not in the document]
penberg: @dekimir I am not sure what you mean. `TypeScript` target was already added way back in 05daaa098290b533662412f2098506ecb3140568. The reason `TypeScript` appears here is because I introduced a _third target_ "index JSON" so we now need code that says "if target is js or ts -> do what we did before; otherwise do the index JSON thing".

chiselc/docs/indexing.md
```diff
@@ -0,0 +1,23 @@
+# Index JSON Reference
+
+The `index-json` target for `chiselc` emits predicate indexes in JSON format.
```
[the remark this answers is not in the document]
penberg: We can't say by static analysis if something _should_ have an index or not. We can only extract predicate properties and give that as a list of _candidate indexes_.

I tweaked the documentation to reflect that.

chiselc/docs/indexing.md
penberg: Sure, fair point, but `filtered-properties.json` is not a good target name to pass to `chiselc` (it's not a _file_).

Some proposals for a better name: `candidate-indexes` and `filter-properties`. Feel free to propose something if you don't like either of them.

server/src/datastore/engine.rs
penberg: @dekimir But won't `index.name()` end up being the _same_ for both cases, which means that `IF NOT EXISTS` prevents a duplicate index from being created? Or did I misunderstand what you're saying here.

dekimir: Yes, it will be the same -- my point is that I don't see how the same index helps in two completely different cases.

chiselc/src/parse.rs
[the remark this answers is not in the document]
dekimir: OK, I see.  But I really dislike `match target` inside `match target`.  Can't you have three `match` branches at the outer level?  Or initialize `module` to its ultimate value on line 87?

server/src/datastore/engine.rs
penberg: Yeah, looks like the index indeed dos not help in the `col1 = col2` case at all:

```
datadb=# CREATE TABLE foo (id INTEGER PRIMARY KEY, col1 INTEGER, col2 INTEGER);
CREATE TABLE
```

Without index:

```
datadb=# EXPLAIN SELECT * FROM foo WHERE col1 = col2;
                      QUERY PLAN                      
------------------------------------------------------
 Seq Scan on foo  (cost=0.00..35.50 rows=10 width=12)
   Filter: (col1 = col2)
(2 rows)

datadb=# EXPLAIN SELECT * FROM foo WHERE col1 > 2 AND col2 < 5;
                      QUERY PLAN                       
-------------------------------------------------------
 Seq Scan on foo  (cost=0.00..40.60 rows=227 width=12)
   Filter: ((col1 > 2) AND (col2 < 5))
(2 rows)

datadb=# 
```


With index:
```
datadb=# CREATE INDEX ON foo (col1, col2);
CREATE INDEX
datadb=# EXPLAIN SELECT * FROM foo WHERE col1 = col2;
                      QUERY PLAN                      
------------------------------------------------------
 Seq Scan on foo  (cost=0.00..35.50 rows=10 width=12)
   Filter: (col1 = col2)
(2 rows)

datadb=# EXPLAIN SELECT * FROM foo WHERE col1 > 2 AND col2 < 5;
                                    QUERY PLAN                                     
-----------------------------------------------------------------------------------
 Bitmap Heap Scan on foo  (cost=11.01..24.41 rows=227 width=12)
   Recheck Cond: ((col1 > 2) AND (col2 < 5))
   ->  Bitmap Index Scan on foo_col1_col2_idx  (cost=0.00..10.95 rows=227 width=0)
         Index Cond: ((col1 > 2) AND (col2 < 5))
(4 rows)
```

But, I don't even know what kind of indexing is needed for the case where the `WHERE` clause filters by columns of the _same table_. 🤔

BearLemma: I'm wondering if multi-column index would be of help here but I'm not even sure how I would implement this kind of search by hand in Rust. I couldn't come up with anything better than linear-time search over all rows which isn't really helpful :-D

penberg: @BearLemma There is a way to index this:

```
datadb=# CREATE INDEX ON foo (col1, col2) WHERE col1 = col2;
datadb=# EXPLAIN SELECT * FROM foo WHERE col1 = col2;
                                    QUERY PLAN                                    
----------------------------------------------------------------------------------
 Bitmap Heap Scan on foo  (cost=4.13..14.69 rows=10 width=12)
   Recheck Cond: (col1 = col2)
   ->  Bitmap Index Scan on foo_col1_col2_idx1  (cost=0.00..4.13 rows=10 width=0)
(3 rows)
```

@dekimir To avoid increasing the scope of this pull request, may I suggest we open a ticket for this improvement and deal with it post-merge?

penberg:
@dekimir Rebased, please re-re-review.

server/src/datastore/engine.rs
BearLemma: hoho, interesting! I've never used conditional indexing, but this makes sense. Nice!

penberg:
Converting to draft mode because there's one critical FIXME to address.

penberg:
Fixed up, please re-re-review @dekimir.

BearLemma:
@penberg I'm not sure if this tells us anything apart from that the index is used (which is great btw). In terms of the speedup it's expected that the speedup is simply going to be something along the lines of `n / log(n)`-times faster with index :man_shrugging: 

Also even in this case, I would argue that the 3ms of "Server Processing" is mostly really just server processing (communication and javascript evaluation and shuffling). So basically the DB execution time went from something to nothing (measurable). 
Which is great :relaxed:

server/src/datastore/engine.rs
dekimir: > To avoid increasing the scope of this pull request, may I suggest we open a ticket for this improvement and deal with it post-merge?

OK

dekimir (commented):
Overall looks good, I just dislike `chiselc` still talking about indices when it's really just enumerating filtered properties.

chiselc/src/indexing.rs
```diff
@@ -0,0 +1,61 @@
+use crate::query::Filter;
+use indexmap::IndexSet;
+use serde::Serialize;
+use swc_ecmascript::ast::{ObjectLit, Prop, PropName, PropOrSpread};
+
+/// An index.
+#[derive(Debug, Serialize)]
+pub struct Index {
```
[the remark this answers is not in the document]
penberg: fixed

chiselc/src/indexing.rs
```diff
@@ -0,0 +1,61 @@
...
+                    _ => {
+                        todo!();
+                    }
+                },
+                _ => {
+                    todo!();
+                }
+            }
+        }
+        Index {
```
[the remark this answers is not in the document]
penberg: added a test and fixed

chiselc/src/rewrite.rs
```diff
@@ -41,22 +44,27 @@ impl FromStr for Target {
...
             _ => Err("Unknown target"),
         }
     }
 }
 
 pub struct Rewriter {
-    target: Target,
     symbols: Symbols,
+    // Accumulated predicate indexes.
+    pub indexes: Vec<FilterProperties>,
```
dekimir, line 56: Not indexes. ;)

chiselc/src/rewrite.rs
```diff
@@ -223,36 +231,36 @@ impl Rewriter {
...
 
-    fn rewrite_call_expr(&self, call_expr: &CallExpr) -> CallExpr {
-        if let Some(filter) = infer_filter(call_expr, &self.symbols) {
-            match self.target {
-                Target::JavaScript | Target::TypeScript => {
-                    return self.to_ts_expr(call_expr, &filter);
-                }
-            }
+    fn rewrite_call_expr(&mut self, call_expr: &CallExpr) -> CallExpr {
+        let (filter, index) = infer_filter(call_expr, &self.symbols);
```
dekimir, line 258: Not index. ;)

proto/chisel.proto
```diff
@@ -72,6 +72,7 @@ message PolicyUpdateRequest {
 
 message ChiselApplyRequest {
    repeated AddTypeRequest types = 1;
+   repeated IndexCandidate index_candidates = 8;
```
dekimir, line 75: Why not put 8 after 7?  Make it easy for the next person extending this to see the high-water mark.

dekimir (approved):
Nitpicks for follow-ups...

PR #1291, opened 2022-05-20
Allow unary expression for default values

### LINKED ISSUE #1171: chiselstrike compiler seems to not allow negative numbers (or null) for typescript defaults
```
import { ChiselEntity, labels } from "@chiselstrike/api"

export class Word extends ChiselEntity {
    day: number = -1;
    word: string = "";
}
```

yields in chiseld console:

```
error: expression not supported
 --> models/word.ts:8:19
  |
8 |     day: number = -1;
  |                   ^^

error: While parsing class Word
 --> models/word.ts:8:5
  |
8 |     day: number = -1;
  |     ^^^^^^^^^^^^^^^^^

Error: expression not supported
```

Changing it to 0 makes it work.

glommer:
We forbid expressions for default values, but we should allow unary
expressions. The reason for that, is that things like "-1", that are
obviously very reasonable defaults, are unary expressions.

Fixes #1171

### DIFF
--- a/cli/src/ts.rs
+++ b/cli/src/ts.rs
@@ -97,6 +97,12 @@ fn get_field_value(handler: &Handler, x: &Option<Box<Expr>>) -> Result<Option<(S
                 let val_type = lit_name(k).into();
                 Ok(Some((val, val_type)))
             }
+            Expr::Unary(k) => {
+                let op = k.op;
+                let value = get_field_value(handler, &Some(k.arg.clone()))?
+                    .ok_or_else(|| swc_err(handler, k, "unexpected empty expression"))?;
+                Ok(Some((format!("{}{}", op, value.0), value.1)))
+            }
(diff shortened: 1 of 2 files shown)


cli/tests/lit/number.deno
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
```
BearLemma, line 1: Can you maybe extend `defaults.deno` ? I would prefer not to create additional lit test files. We already have so many it's very hard to navigate.

glommer: kk

glommer:
@BearLemma updated.

cli/tests/lit/number.deno
glommer: done now

BearLemma (approved):

PR #1293, opened 2022-05-20
Implement findOne with predicate.

BearLemma:
Extends `findOne` to support filtering with predicate lambda.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -673,14 +673,50 @@ export class ChiselEntity {
         return await it.toArray();
     }
 
+    /**
+     * Returns a single object of type T for which the given `predicate` returns true.
+     *
+     * If more than one match is found, any is returned
+     */
+    static async findOne<T extends ChiselEntity>(
+        this: { new (): T },
+        predicate: (arg: T) => boolean,
+    ): Promise<T | undefined>;
+
     /** Returns a single object that matches the `Partial` object `restrictions` passed as its parameter.
      *
      * If more than one match is found, any is returned. */
     static async findOne<T extends ChiselEntity>(
         this: { new (): T },
         restrictions: Partial<T>,
+    ): Promise<T | undefined>;
+
+    static async findOne<T extends ChiselEntity>(
+        this: { new (): T },
+        arg1: ((arg: T) => boolean) | Partial<T>,
     ): Promise<T | undefined> {
-        const it = chiselIterator<T>(this).filter(restrictions).take(1);
+        let it = undefined;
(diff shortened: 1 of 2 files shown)


glommer (approved):

api/src/chisel.ts
```diff
@@ -673,14 +673,50 @@ export class ChiselEntity {
...
         this: { new (): T },
         restrictions: Partial<T>,
+    ): Promise<T | undefined>;
+
+    static async findOne<T extends ChiselEntity>(
+        this: { new (): T },
+        arg1: ((arg: T) => boolean) | Partial<T>,
+    ): Promise<T | undefined> {
+        let it = undefined;
+        if (typeof arg1 == "function") {
```
dekimir, line 699: @barbieri this `if` is necessitated by TypeScript's overload-resolution rules. Note that the `then` and `else` branches are identical. Is there some trick to avoid it?

barbieri: I need the error to say for sure, but given the code is identical there should be a way. Could you paste the whole TS error and context here so I can try to see what's happening?



Nitpick: it should be triple `===` in JS (https://eslint.org/docs/rules/eqeqeq)

dekimir: If I replace the whole `if-else` block with just the `it = chiselIterator<T>(this).filter(arg1);` from line 700:
```
  Error: Compilation failed:
  src/chisel.ts:699:49 - error TS2769: No overload matches this call.
    Overload 1 of 2, '(predicate: (arg: T) => boolean): ChiselCursor<T>', gave the following error.
      Argument of type '((arg: T) => boolean) | Partial<T>' is not assignable to parameter of type '(arg: T) => boolean'.
        Type 'Partial<T>' is not assignable to type '(arg: T) => boolean'.
          Type 'Partial<T>' provides no match for the signature '(arg: T): boolean'.
    Overload 2 of 2, '(restrictions: Partial<T>): ChiselCursor<T>', gave the following error.
      Argument of type '((arg: T) => boolean) | Partial<T>' is not assignable to parameter of type 'Partial<T>'.
        Type '(arg: T) => boolean' has no properties in common with type 'Partial<T>'.

  699             it = chiselIterator<T>(this).filter(arg1);
                                                      ~~~~
```

barbieri: got it... as stupid as it may look, it seems you may need to add another version that takes both... sometimes I really don't get the TS folks :-/ 

Alternatively you can force a variant with `as Partial<T>`, not ideal, but since the returns are the same, it would silent the tsc without adding extra runtime overhead.

PR #1294, opened 2022-05-20
Use extends ChiselEntity in multiple places in chisel.ts.

BearLemma:
We already do `<T extends ChiselEntity>` in multiple places, this makes it slightly more coherent.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -517,7 +517,7 @@ export class ChiselRequest extends Request {
     }
 }
 
-export function chiselIterator<T>(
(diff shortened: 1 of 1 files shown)


dekimir (approved):
How can this be tested?

BearLemma:
> How can this be tested?

I don't think it can be tested :thinking:  Or at least can't think of a way to do it. We can't have check that all static methods have a template arg that extends ChiselEntity as there can be other methods that don't by design.
I this case, I think we just need to remember to do it :man_shrugging:

PR #1295, opened 2022-05-23
Add the endpoint_compile crate

espindola:
This small crate refactors the special logic for compiling endpoints,
as opposed to plain ts files, to a new crate. This cannot be part of
tsc_compile sice tsc_compile is used to compile api which is used by
the endpoints.

With this my_tsc complies endpoints the same way that "chisel apply"
does.

### DIFF
--- a/endpoint_tsc/src/lib.rs
+++ b/endpoint_tsc/src/lib.rs
@@ -1,0 +1,51 @@
+// SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
+
+use anyhow::Result;
+use std::collections::HashMap;
+use std::io::Write;
+use tempfile::Builder;
+use tempfile::NamedTempFile;
+use tsc_compile::CompileOptions;
+
+pub struct Compiler {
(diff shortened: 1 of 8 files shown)


endpoint_tsc/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+impl Compiler {
+    pub fn new(use_snapshot: bool) -> Compiler {
+        let tsc = tsc_compile::Compiler::new(use_snapshot);
+        Compiler { tsc }
+    }
+
+    pub async fn compile_endpoint(&mut self, file_name: &str) -> Result<HashMap<String, String>> {
+        let mods: HashMap<String, String> = [(
```
BearLemma, line 21: Not really related to this PR, but it might be a bit more elegant to do 
`let mods = HashMap::from([(.., ..)]);

BearLemma: yup, certainly.

BearLemma (approved):

PR #1305, opened 2022-05-24
Add tests

espindola:
This at least documents what errors we currently produce. I will improve the location in the error message in the next PR.


espindola:
> What are you testing? My understanding is that you're testing if we're properly calling the TS compiler on the given code and propagating any errors from it. Is that right?

Mostly. I added comments to the tests themselves, but they are testing that we correctly report the error locations (or not).

athoscouto (approved):

PR #1307, opened 2022-05-24
Refactor url insertion code

espindola:
I want to insert urls in the download map from other places, so
refactor it.

### DIFF
--- a/tsc_compile/src/lib.rs
+++ b/tsc_compile/src/lib.rs
@@ -53,9 +53,19 @@ impl DownloadMap {
     fn len(&self) -> usize {
         self.path_to_url.len()
     }
-    fn insert(&mut self, path: String, url: Url) {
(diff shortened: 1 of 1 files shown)


tsc_compile/src/lib.rs
```diff
@@ -53,9 +53,17 @@ impl DownloadMap {
     fn len(&self) -> usize {
         self.path_to_url.len()
     }
-    fn insert(&mut self, path: String, url: Url) {
+    fn insert(&mut self, url: Url) -> String {
```
BearLemma, line 56: It would be nice if this function had a blurb saying what this function actually does, I'm a little confused what's going on here.

espindola: added.

BearLemma (approved):

PR #1310, opened 2022-05-24
Make without_extension more specific

espindola:
We really just want to remove the JS/TS extensions. In the current
code base both implementations work, but this change makes
without_extension more reliable. For example, given
https://cdn.skypack.dev/handlebars, it used to return
https://cdn.skypack, now it returns the original url.

### DIFF
--- a/tsc_compile/src/lib.rs
+++ b/tsc_compile/src/lib.rs
@@ -234,7 +234,12 @@ fn abs(path: &str) -> String {
 }
 
 fn without_extension(path: &str) -> &str {
-    path.rsplit_once('.').map_or(path, |p| p.0)
+    for suffix in [".d.ts", ".ts", ".js"] {
+        if let Some(s) = path.strip_suffix(suffix) {
+            return s;
+        }
+    }
+    path
 }
 
 static SNAPSHOT: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/SNAPSHOT.bin"));


tsc_compile/src/lib.rs
```diff
@@ -234,7 +234,12 @@ fn abs(path: &str) -> String {
 }
 
 fn without_extension(path: &str) -> &str {
-    path.rsplit_once('.').map_or(path, |p| p.0)
+    for suffix in [".d.ts", ".ts", ".js"] {
```
BearLemma, line 237: As you have provided nice example of what didn't want previous, would you mind adding a unit test? This looks very testable.

espindola: It is testable, but it is such an implementation detail that it feels like a case where I would be testing that copy and paste is working :-)

espindola: For instance, if there was a typo in there with x.ts instead of d.ts, we would get real test failures, and I would probably copy and past the x.ts into the unit test anyway.

penberg: The function is now indeed so simple that an unit test doesn't add much value.

However, a counter-argument is that as the function had a bug, we should consider adding a test to catch any future regressions. IOW, there's a strong case for checking that `https://cdn.skypack.dev/handlebars` returns the right value.

BearLemma: Yeah, that's sort of in the spirit of what I meant, I agree that it's very simplistic, but something should have caught that problem, be it a very specialized unit test or something up the pipe.

espindola: There is no bug. The function works correctly for what it is used. A followup PR will want to use in more cases. I extracted this PR since the new version works on both and it is one less change to be reviewed later.

penberg (approved):

PR #1313, opened 2022-05-25
Make chiselc transform find{Many,One} calls

penberg:
This extends `chiselc` to also transform `findMany()` and `findOne()`
with predicate lambdas to internal query calls that are optimized by the
runtime.

Similarly, `chiselc` now also extracts filter properties from `findMany`
and `findOne`, which allows the runtime to auto-index the underlying
table columns if needed.

### DIFF
--- a/chiselc/src/transforms/find.rs
+++ b/chiselc/src/transforms/find.rs
@@ -1,0 +1,29 @@
+use crate::filtering::FilterProperties;
+use crate::query::Operator as QOperator;
+use crate::symbols::Symbols;
+use crate::transforms::utils::{extract_filter, lookup_callee_entity_type};
+use crate::utils::is_call_to_entity_method;
+
+use swc_ecmascript::ast::{CallExpr, Callee};
+
+/// Infer filter operator from the lambda predicate of `findMany`.
+pub fn infer_find(
+    call_expr: &CallExpr,
+    symbols: &Symbols,
+) -> (Option<Box<QOperator>>, Option<FilterProperties>) {
+    if !is_rewritable_find("findMany", &call_expr.callee, symbols) {
+        return (None, None);
+    }
+    let entity_type = match lookup_callee_entity_type(&call_expr.callee) {
+        Ok(entity_type) => entity_type,
+        _ => return (None, None),
+    };
+    extract_filter(call_expr, entity_type, "__findMany".to_string())
+}
+
+fn is_rewritable_find(function: &str, callee: &Callee, symbols: &Symbols) -> bool {
+    match callee {
+        Callee::Expr(expr) => is_call_to_entity_method(expr, function, symbols),
+        _ => false,
+    }
+}
--- a/chiselc/src/transforms/find.rs
+++ b/chiselc/src/transforms/find.rs
@@ -6,7 +6,7 @@ use crate::utils::is_call_to_entity_method;
 
 use swc_ecmascript::ast::{CallExpr, Callee};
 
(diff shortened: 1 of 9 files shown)


penberg:
In draft mode because depends on https://github.com/chiselstrike/chiselstrike/pull/1290

chiselc/tests/lit/filter-properties-findmany.lit
```diff
@@ -0,0 +1,18 @@
+// Extract predicate indexes
+// RUN: @chiselc @file -e Person --target filter-properties
+
+await Person.findMany({ name: "Pekka", age: 39 });
```
BearLemma, line 4: Shouldn't there be some checks or something? Or is this testing just that the compiler won't die?

chiselc/src/transforms/find.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::query::Operator as QOperator;
+use crate::query::Scan as QScan;
+use crate::symbols::Symbols;
+use crate::transforms::utils::{convert_bin_expr, lookup_callee_entity_type};
+use crate::utils::{is_call_to_entity_method, pat_to_string};
+
+use swc_ecmascript::ast::{BlockStmtOrExpr, CallExpr, Callee, Expr, Lit, Stmt};
+
+/// Infer filter operator from the lambda predicate of `findMany`.
+pub fn infer_find(
```
BearLemma, line 14: Could you maybe split this function into smaller function blocks? 
I think it should be possible to at least separate the logic that makes the decision to transform or not to transform and a second part that does the transformation itself.

chiselc/src/transforms/find.rs
```diff
@@ -0,0 +1,107 @@
...
+    assert_eq!(args.len(), 1);
+    let arg = &args[0];
+    let arrow = match &*arg.expr {
+        Expr::Arrow(arrow_expr) => arrow_expr,
+        Expr::Object(object_lit) => {
+            /*
+             * Filter by restriction object, nothing to transform, but let's
+             * grab predicate indexes.
+             */
+            let props = FilterProperties::from_object_lit(entity_type, object_lit);
```
BearLemma, line 36: Plus I sort of don't like that it's trying to do both transformation and index detection at once... I understand that it's more efficient and maybe easier to write, but it's also harder to keep mental track of what's going on.

chiselc/tests/lit/filter-properties-findmany.lit
penberg: @BearLemma All the checks are at the bottom of the `lit` file because `chiselc` generates a single-line JSON for the whole TypeScript file (that the `lit` file represents).

chiselc/src/transforms/find.rs
penberg: @BearLemma I agree it's not ideal, but please note that this is exactly what `filter.rs` is already doing, not something new to this pull request.

Also, I am hesitant to change this now, because https://github.com/chiselstrike/chiselstrike/issues/1377 suggests that the JSON format is not sufficient, and the runtime should perhaps just use the query expression to index-generation.

chiselc/src/transforms/find.rs
penberg: `infer_find` calls `infer_rewritable_find` to make the decision to transform, but I guess the part where it builds `query::Operator` could be extracted to a separate function.

penberg: @BearLemma Fixed

PR #1315, opened 2022-05-25
Add a test

espindola:
This shows that we are not outputting imported files. It will be fixed
in a followup patch.


athoscouto (commented):
I think you got a typo.
`s/output_imorted/output_imported/g`

athoscouto (approved):

PR #1320, opened 2022-05-26
Simplify our TSC integration

espindola:
This includes two patches that should have no functionality change, they just simplify the code.

### DIFF
--- a/tsc_compile/src/lib.rs
+++ b/tsc_compile/src/lib.rs
@@ -331,7 +331,7 @@ impl Compiler {
         file_name: &str,
         opts: CompileOptions<'_>,
     ) -> Result<HashMap<String, String>> {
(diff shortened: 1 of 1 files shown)


tsc_compile/src/lib.rs
```diff
@@ -383,10 +345,13 @@ impl Compiler {
         let mut loader = ModuleLoader { extra_libs };
         let resolver = ModuleResolver { extra_libs: to_url };
 
-        let maybe_imports = if let Some(path) = opts.extra_default_lib {
+        let extra_default_lib = opts
+            .extra_default_lib
+            .map(|path| "file://".to_string() + &abs(path));
```
BearLemma, line 350: maybe `format("file://{}", abs(path))` would be nicer?

BearLemma (approved):

PR #1321, opened 2022-05-26
Add missing format string

espindola:
The first argument to printf is a format string. Passing a single
string is equivalent only if that string has no formatting directives.


BearLemma:
@espindola if we are to do it in a separate commit, would you mind fixing it in `defaults.deno` as well? I've used the same technique there.

espindola:
> @espindola if we are to do it in a separate commit, would you mind fixing it in `defaults.deno` as well? I've used the same technique there.

Done. Thanks.

BearLemma (approved):

PR #1323, opened 2022-05-26
Implement minBy and maxBy TS-side aggregations.

BearLemma:
This is just a TypeScript-side min/max implementation which will be improved later with the propagation of aggregations to backend. It proved to be rather challenging so I've decided to do it in two steps and first have something that works and is testable and improve upon it later.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -389,6 +389,48 @@ export class ChiselCursor<T> {
         );
     }
 
+    /**
+     * Finds minimal value over all elements using their `key` attribute.
+     *
+     * @param key specifies which attribute of `T` is to be used to chose the minimum.
+     * @returns minimal value of attribute called `key` across all elements. Undefined
+     * values are ignored. If there are no elements or non-undefined values,
+     * the function returns undefined.
+     */
+    async minBy<K extends keyof T>(key: K): Promise<T[K] | undefined> {
+        let min = undefined;
+        for await (const e of this) {
+            const val = e[key];
+            if (min === undefined) {
+                min = val;
+            } else if (val !== undefined && val < min) {
+                min = val;
+            }
+        }
+        return min;
+    }
+
+    /**
+     * Finds maximal value over all elements using their `key` attribute.
+     *
+     * @param key specifies which attribute of `T` is to be used to chose the maximum.
+     * @returns maximal value of attribute called `key` across all elements. Undefined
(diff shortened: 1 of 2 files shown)


api/src/chisel.ts
```diff
@@ -389,6 +389,48 @@ export class ChiselCursor<T> {
...
 
+    /**
+     * Finds minimal value over all elements using their `key` attribute.
+     *
+     * @param key specifies which attribute of `T` is to be used to chose the minimum.
+     * @returns minimal value of attribute called `key` across all elements. Undefined
+     * values are ignored. If there are no elements or non-undefined values,
+     * the function returns undefined.
+     */
+    async minBy<K extends keyof T>(key: K): Promise<T[K] | undefined> {
```
espindola, line 400: Ok, so we basically map SQL NULL to undefined? Wouldn't null be a closer match? I remember a similar discussion somewhere, but unfortunately don't remember the decision.

api/src/chisel.ts
```diff
@@ -389,6 +389,48 @@ export class ChiselCursor<T> {
...
+
+    /**
+     * Finds maximal value over all elements using their `key` attribute.
+     *
+     * @param key specifies which attribute of `T` is to be used to chose the maximum.
+     * @returns maximal value of attribute called `key` across all elements. Undefined
+     * values are ignored. If there are no elements or non-undefined values,
+     * the function returns undefined.
+     */
+    async maxBy<K extends keyof T>(key: K): Promise<T[K] | undefined> {
```
espindola, line 421: We will get more aggregates than min and max, right? Would it be worth it coding a fold function that handle the undefined case and only calls the callback with two defined values?

espindola (commented):
I think this is OK. Any particular reason to not include a SQL implementation too? Other than the comment about null X undefined this seems to match the SQL semantics, but an implementation would be good to make sure.

api/src/chisel.ts
BearLemma: Yeah the decision was to use undefined. `findOne` was the example.

espindola (approved):
I would still suggest adding a fold function and using that to implement min and max, but that can be a followup.

BearLemma:
> I think this is OK. Any particular reason to not include a SQL implementation too? Other than the comment about null X undefined this seems to match the SQL semantics, but an implementation would be good to make sure.

The reason is that to propagate the thing to SQL takes surprising amount of work.. I'm doing that right now

PR #1325, opened 2022-05-26
Use jq

espindola:
This makes it clear that the output is json and what we are looking
for.


BearLemma:
@espindola It would be good to fix also some of the other occurrences of JSON parsing in this file as well... :thinking:

espindola:
> @espindola It would be good to fix also some of the other occurrences of JSON parsing in this file as well... thinking

Any easy way to find them all?

BearLemma (approved):

PR #1327, opened 2022-05-26
Upgrade lit to include always_show_stdout flag.

BearLemma:
Showing the whole stdout is very useful for debugging and the current version of lit doesn't show stdout if anything is printed to stderr. :(

### DIFF
--- a/cli/tests/integration_tests.rs
+++ b/cli/tests/integration_tests.rs
@@ -159,6 +159,7 @@ fn run_tests(opt: Opt, optimize: bool) -> bool {
                     let i = rayon::current_thread_index().unwrap_or(0) + 1;
                     config.test_paths = vec![test_path.clone()];
                     config.truncate_output_context_to_number_of_lines = Some(500);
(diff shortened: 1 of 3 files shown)


cli/tests/integration_tests.rs
```diff
@@ -159,6 +159,7 @@ fn run_tests(opt: Opt, optimize: bool) -> bool {
                     let i = rayon::current_thread_index().unwrap_or(0) + 1;
                     config.test_paths = vec![test_path.clone()];
                     config.truncate_output_context_to_number_of_lines = Some(500);
+                    config.always_show_stdout = false;
```
espindola, line 162: This is the default value.

athoscouto (approved):

cli/tests/integration_tests.rs
BearLemma: It is but I wanted to have it there so that I don't always have to look how is the thing named. I switch it on quite often lately.

PR #1328, opened 2022-05-26
Don't map url to paths

espindola:
This is a bit of a hack as it requires dropping noEmitOnError and
working around TSC not wanting to override a .js.

I still think it is worth it, in part because it is a lot less code,
in part because it is closer to how deno calls TSC.

### DIFF
--- a/tsc_compile/src/lib.rs
+++ b/tsc_compile/src/lib.rs
@@ -18,6 +18,7 @@ use deno_graph::source::LoadResult;
 use deno_graph::source::Loader;
 use deno_graph::source::ResolveResponse;
 use deno_graph::source::Resolver;
+use deno_graph::MediaType;
 use deno_graph::ModuleGraph;
 use deno_graph::ModuleKind;
 use serde::de::DeserializeOwned;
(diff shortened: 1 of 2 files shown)


tsc_compile/src/lib.rs
```diff
@@ -392,26 +345,29 @@ impl Compiler {
...
         let map = op_state.take::<DownloadMap>();
-        if ok {
+        if map.diagnostics.is_empty() {
             let mut prefix_map: HashMap<&str, &Url> = HashMap::default();
-            for (k, v) in &map.path_to_url {
-                prefix_map.insert(without_extension(k), v);
-            }
             let mut ret: HashMap<String, String> = HashMap::default();
+            for m in map.graph.modules() {
+                let u = &m.specifier;
```
BearLemma, line 358: maybe a bit more descriptive variable name would be appropriate, I have no idea what u should stand for.

espindola:
> > noEmitOnError
> 
> Reading upon the noEmitOnError, I don't even know when could I use the compiler output file if any compilation errors are detected. Is it for a situation where the TS compiler detects errors, but the underlying javascript can still be runable?

Yes. The one case we consider that that is the case is the one we filter in

```
        allDiagnostics = allDiagnostics.filter(({code}) => {
            // Ignore the error
            //   Cannot write file '...' because it would overwrite input file.
            // It happens when trying to compile an URL that doesn't end in .ts
            // as TSC thinks it will be overwriting the input.
            return code != 5055;
        });
```

That is, when TSC produces an error because it thinks it would be overwriting `http://example.com/foo.js` with `http://example.com/foo.js`.

BearLemma (approved):

PR #1336, opened 2022-05-30
cli/apply: Fix chiselc in Node mode

### LINKED ISSUE #1317: Apply fails to transform endpoint code in Node mode
If you run `chisel apply` in Node mode, `chiselc` is invoked, but on the wrong file. That is, `chiselc` is run on a "bundler entry" file, which is generated file, but not on the actual endpoint code that the "bundler entry" imports, which means filter by lambda is not getting transformed.

penberg:
Fixes #1317


dekimir (approved):
How do we test this?

penberg:
@dekimir For functional testing, we just run our test suite with the optimizer on and check that the tests pass. Of course, they also pass if the optimizer does nothing, which is why I am working on a performance harness. That's how I discovered that the indexing stuff, for example, was not actually getting applied.

We could add a test case checking that `chisel apply` generates the transformed endpoints in the `.gen` directory, but I am not sure we want to test that implementation detail because you could imagine `chisel apply` not generating them and, instead, performing the transformation purely in memory.

penberg:
@dekimir Looks like I missed this post-merge comment. I am not sure what kind of test case you are thinking of here. The issue was that `chisel apply` was invoking `chiselc` on the wrong file (`chisel apply` internally generates a wrapper endpoint) so `chiselc/tests/lit` plays no part in testing.

It is also unclear what kind of test case we could add to `cli/test/lit` because the whole point of running `chiselc` as part of `chisel apply` is that it's transparent and the behavior of the endpoints stay the same (except for better performance). I mean, we can, of course, check for the presence of a `.gen/<endpoint>` file (and even verify contents), but that's an implementation detail that may change in the future.

PR #1337, opened 2022-05-30
Expand tests

espindola:
With this we now also test import from sibling directories and test variations on the path of the given source file. The output will change a bit in my next PR, so this serves to make that change more visible.

### DIFF
--- a/tsc_compile/src/lib.rs
+++ b/tsc_compile/src/lib.rs
@@ -417,6 +417,7 @@ mod tests {
     use super::CompileOptions;
     use anyhow::Result;
     use deno_core::anyhow;
+    use std::future::Future;
     use std::io::Write;
     use tempfile::Builder;
     use tempfile::NamedTempFile;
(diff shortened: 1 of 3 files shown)


tsc_compile/src/lib.rs
```diff
@@ -164,11 +166,24 @@ where
     Ok(res)
 }
 
+// This is similar to PathBuf::push, but folds "/./" and "/../". This
+// is similar to what TSC own path handling does.
+fn join_path(mut base: PathBuf, rel: &Path) -> PathBuf {
```
BearLemma, line 171: From the blurb I'm unable to gather what is this function supposed to do :thinking:  Isn't it equivalent to standard PathBuf join and consequent application of https://doc.rust-lang.org/std/fs/fn.canonicalize.html ?

tsc_compile/src/lib.rs
```diff
@@ -739,4 +739,25 @@ export default foo;
         check_import_js("./tests/import_js_a.ts").await;
         check_import_js(&abs("tests/import_js_a.ts")).await;
     }
+
+    async fn check_relative(path: &str) {
+        let abs_a = abs(path);
+        let import = format!(
```
BearLemma: Might be better to pass the prefix path into the test function directly so that you don't have to do this juggling. I think it would help readability.

tsc_compile/src/lib.rs
```diff
@@ -689,27 +704,39 @@ export default foo;
...
-        let import = format!("file://{}b.js", path.strip_suffix("a.ts").unwrap());
+    async fn output_imported() {
+        check_output_imported("tests/output_imported_a.ts").await;
+        check_output_imported("./tests/output_imported_a.ts").await;
+        check_output_imported(&abs("tests/output_imported_a.ts")).await;
+    }
+
+    async fn check_import_js(path: &str) {
+        let abs_a = abs(path);
+        let import = format!("file://{}b.js", abs_a.strip_suffix("a.ts").unwrap());
```
BearLemma: Similar here, I would just pass the prefix path and joined a's and b's paths in the function.

BearLemma:
The comments I made were for the first two commits which get changed quite a bit by the later two, but I think they still apply in essence.

tsc_compile/src/lib.rs
espindola: No. Canonicalize will resolve symbolic links. In fact, on unix it just calls realpath.

BearLemma (approved):

PR #1338, opened 2022-05-30
Add aggregation operator.

BearLemma:
This PR adds aggregation operator used by `minBy` and `maxBy`. In consequent PRs, this operator will be shipped to our backend and evaluated directly in the database.
It's composed of three separate patches, so it might be better to go one by one.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -347,6 +347,87 @@ class SortBy<T> extends Operator<T, T> {
     }
 }
 
+/**
+ * AggregateBy operator is an intermediate Operator used to implement various aggregation
+ * operators like MinBy/MaxBy. It provides a general aggregate interface performing a fold
+ * along `Input[K]` field values using `init` initial value and aggerage (fold) operator
+ * `aggregateOp`.
+ *
+ * This Operator can't be used directly as our Rust backend uses the name of the operator
+ * for identification.
+ */
+abstract class AggregateBy<Input, K extends keyof Input, Output>
+    extends Operator<Input, Output> {
+    constructor(
+        inner: Operator<unknown, Input>,
+        private key: K,
+        private init: Output,
+        private aggregateOp: (lhs: Output, rhs: Input[K]) => Output,
+    ) {
+        super(inner);
+    }
+    apply(
+        iter: AsyncIterable<Input>,
+    ): AsyncIterable<Output> {
+        const key = this.key;
+        const init = this.init;
+        const aggregateOp = this.aggregateOp;
+        return {
+            [Symbol.asyncIterator]: async function* () {
+                let result = init;
+                for await (const e of iter) {
+                    result = aggregateOp(result, e[key]);
+                }
+                yield result;
+            },
+        };
+    }
+
+    public eval(): AsyncIterable<Output> {
+        let iter = this.inner!.eval();
+        if (iter === undefined) {
+            iter = this.inner!.runChiselQuery();
+        }
+        return this.apply(iter);
+    }
+
+    recordToOutput(rawRecord: unknown): Output {
+        return rawRecord as Output;
+    }
+}
+
+class MinBy<Input, K extends keyof Input>
+    extends AggregateBy<Input, K, Input[K] | undefined> {
+    constructor(inner: Operator<unknown, Input>, key: K) {
+        const min = (v1: Input[K] | undefined, v2: Input[K]) => {
+            if (v2 === undefined || v2 === null) {
+                return v1;
+            }
+            if (v1 === undefined) {
+                return v2;
+            }
+            return v1 < v2 ? v1 : v2;
+        };
+        super(inner, key, undefined, min);
+    }
+}
+
+class MaxBy<Input, K extends keyof Input>
+    extends AggregateBy<Input, K, Input[K] | undefined> {
+    constructor(inner: Operator<unknown, Input>, key: K) {
+        const max = (v1: Input[K] | undefined, v2: Input[K]) => {
+            if (v2 === undefined || v2 === null) {
+                return v1;
+            }
+            if (v1 === undefined) {
+                return v2;
+            }
+            return v1 > v2 ? v1 : v2;
+        };
+        super(inner, key, undefined, max);
+    }
+}
+
 /** ChiselCursor is a lazy iterator that will be used by ChiselStrike to construct an optimized query. */
 export class ChiselCursor<T> {
     constructor(private inner: Operator<unknown, T>) {}
@@ -473,16 +554,12 @@ export class ChiselCursor<T> {
      * the function returns undefined.
      */
     async minBy<K extends keyof T>(key: K): Promise<T[K] | undefined> {
-        let min = undefined;
-        for await (const e of this) {
-            const val = e[key];
-            if (min === undefined) {
-                min = val;
-            } else if (val !== undefined && val < min) {
-                min = val;
-            }
+        const c = new ChiselCursor(
+            new MinBy<T, K>(this.inner, key),
+        );
+        for await (const min of c) {
+            return min;
         }
-        return min;
     }
 
     /**
@@ -494,16 +571,12 @@ export class ChiselCursor<T> {
      * the function returns undefined.
      */
     async maxBy<K extends keyof T>(key: K): Promise<T[K] | undefined> {
-        let max = undefined;
-        for await (const e of this) {
-            const val = e[key];
-            if (max === undefined) {
-                max = val;
-            } else if (val !== undefined && val > max) {
-                max = val;
-            }
+        const c = new ChiselCursor(
+            new MaxBy<T, K>(this.inner, key),
+        );
+        for await (const max of c) {
+            return max;
         }
-        return max;
     }
 
     /** Executes the function `func` for each element of this cursor. */


api/src/chisel.ts
```diff
@@ -452,20 +490,31 @@ export class ChiselCursor<T> {
...
-     * all subsequent operations are applied on the resulting async iterable in
-     * TypeScript. In such a case, the function returns the resulting AsyncIterable.
+     * chain. If PredicateFilter or aggregation is encountered, a backend query is
+     * generated and all subsequent operations are applied on the resulting async
+     * iterable in TypeScript. In such a case, the function returns the resulting
+     * AsyncIterable.
      * If no PredicateFilter is found, undefined is returned.
      */
-    private makeTransformedQueryIter(
+    private _makeTransQueryIter(
```
penberg, line 517: Can you please just keep the `makeTransformedQueryIter` naming (this would be one with underscore, though). It wasn't obvious to me that the "trans" abbrevation refers to _transform_ when I was reading this. Furthermore, it's always better to _rename_ stuff independently of changing the logic.

api/src/chisel.ts
```diff
@@ -419,16 +454,19 @@ export class ChiselCursor<T> {
...
-                max = val;
-            } else if (val !== undefined && val > max) {
-                max = val;
-            }
+        const opChain = new AggregateBy<T, K>(
+            key,
+            this.inner,
+            OpType.MaxBy,
+            (v1, v2) => {
+                return v1 > v2 ? v1 : v2;
```
penberg, line 462: Purely style issue but perhaps move this lambda into a local variable and use it here for readability?

api/src/chisel.ts
```diff
@@ -419,16 +454,19 @@ export class ChiselCursor<T> {
...
+            },
+        );
+        const iter = this.makeTransQueryIter(opChain);
+        for await (const max of iter) {
+            return max as unknown as T[K];
```
penberg, line 467: I don't understand this code: why do we know that the iterator only has _one_ element? Shouldn't we have a sanity check for that? Or use something else than an iterator for the interface?

api/src/chisel.ts
```diff
@@ -475,12 +524,15 @@ export class ChiselCursor<T> {
...
             );
         }
-        let iter = this.makeTransformedQueryIter(op.inner);
+        let iter = this._makeTransQueryIter(op.inner);
         if (iter !== undefined) {
-            return op.apply(iter);
-        } else if (op.type == OpType.PredicateFilter) {
+            return op.apply(iter) as AsyncIterable<T>;
+        } else if (
+            op.type == OpType.PredicateFilter || op.type == OpType.MinBy ||
```
penberg, line 531: Let's have one equals expression per line for readability, please.

api/src/chisel.ts
```diff
@@ -23,26 +25,23 @@ abstract class Operator<T> {
...
-    public containsType(opType: OpType): boolean {
-        if (this.type == opType) {
-            return true;
-        } else if (this.inner === undefined) {
-            return false;
-        } else {
-            return this.inner.containsType(opType);
-        }
+    public useConstructor(): boolean {
+        return this._useConstructor &&
```
penberg, line 43: Why does the property name start with an underscore?

api/src/chisel.ts
BearLemma: Well first of all, my hands are a little bit tight here as well due to the typing problems I've described in the PR's description. 
We know this iterator has one or no elements because It's the result of the `AggregateBy` operator which is currently implemented so that it returns only one operator. Admittedly, it's not great and my original idea was to make the Operator explicitly return just one element, but it's a TypeScript nightmare :(

api/src/chisel.ts
BearLemma: I was lazy to come up with a name that wouldn't conflict with the method's name. Renamed to usesConstructor.

api/src/chisel.ts
```diff
@@ -32,17 +33,13 @@ abstract class Operator<T> {
...
-        if (this.type == opType) {
-            return true;
-        } else if (this.inner === undefined) {
-            return false;
-        } else {
-            return this.inner.containsType(opType);
-        }
+    public useConstructor(): boolean {
+        return this.usesConstructor &&
+            (this.inner == undefined || this.inner.useConstructor());
```
espindola, line 44: ===

BearLemma: Fixed.

api/src/chisel.ts
```diff
@@ -452,11 +452,21 @@ export class ChiselCursor<T> {
...
     [Symbol.asyncIterator](): AsyncIterator<T> {
-        let iter = this.makeTransformedQueryIter(this.inner);
+        const iter = this.makeTransformedQueryIter(this.inner);
+        return iter[Symbol.asyncIterator]();
+    }
+
+    /** Makes asynchronous iterable from given operator. For further documentation,
+     * please refer to the docs of `_makeTransformedQueryIter`.
+     */
+    private makeTransformedQueryIter(
```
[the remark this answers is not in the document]
BearLemma: Ahhhhhh, that will make it even longer. We can surely figure out some better name, this is terrible. How about `foo1`?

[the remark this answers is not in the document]
espindola: It dose make an Iterable. If you find the name too long, makeQueryIterable is fine too, but reusing the name of another function is confusing and makes for a noisy patch, not to mention the name starting with a _.

api/src/chisel.ts
```diff
@@ -264,6 +261,56 @@ class SortBy<T> extends Operator<T> {
...
+                }
+            },
+        };
+    }
+}
+
+class MinBy<T, K extends keyof T> extends AggregateBy<T, K> {
+    constructor(key: K, inner: Operator<T>) {
+        const min = (v1: T[K], v2: T[K]) => {
+            return v1 < v2 ? v1 : v2;
```
dekimir, line 299: Do we have a spec on how NULLs should be treated?  The JavaScript rules will lead to some surprising and inconsistent results.

BearLemma: I wanted to brush it off that I'm already handling that case, but it seems that I've somehow lost the code that handled it :-D Will re-add it to make it aligned with how the Postgres implementation works.

BearLemma: Fixed.

api/src/chisel.ts
```diff
@@ -264,6 +261,59 @@ class SortBy<T> extends Operator<T> {
...
+        iter: AsyncIterable<T>,
+    ): AsyncIterable<T[K]> {
+        const key = this.key;
+        const aggregateOp = this.aggregateOp;
+        return {
+            [Symbol.asyncIterator]: async function* () {
+                let result = undefined;
+                for await (const e of iter) {
+                    const val = e[key];
+                    if (val === undefined || val === null) {
```
dekimir, line 282: I don't think this belongs here.  What if the aggregation function isn't a comparator but something that counts on seeing the NULLs?

BearLemma: Ok, changed the aggregation operator to be more future proof, but to be honest, I don't think it's a much of an improvement and I think we could have done it later when we need more generality.

api/src/chisel.ts
```diff
@@ -12,6 +12,7 @@ abstract class Operator<T> {
     readonly type;
     constructor(
         public readonly inner: Operator<T> | undefined,
+        private readonly usesConstructor: boolean,
```
dekimir, line 15: This is complex and error-prone.  Why not define `useConstructor()` to return `true` here and `false` in `ColumnsSelect`?

BearLemma: Fixed.

BearLemma: At least I think I got what you meant. Please check it out and let me know if the current version looks better :]

penberg:
@BearLemma @dekimir Let's move this forward so that it doesn't bit-rot.

api/src/chisel.ts
```diff
@@ -347,6 +347,87 @@ class SortBy<T> extends Operator<T, T> {
...
+                }
+                yield result;
+            },
+        };
+    }
+
+    public eval(): AsyncIterable<Output> {
+        let iter = this.inner!.eval();
+        if (iter === undefined) {
+            iter = this.inner!.runChiselQuery();
```
espindola, line 389: So, this forces the aggregation to happen is JS. Is this something that will change in another PR? Or this is just fallback and derived classes are expected to implement eval? If the second, it would probably be better to not implement eval in here and implement it in the derived classes.

api/src/chisel.ts
```diff
@@ -347,6 +347,87 @@ class SortBy<T> extends Operator<T, T> {
...
+    recordToOutput(rawRecord: unknown): Output {
+        return rawRecord as Output;
+    }
+}
+
+class MinBy<Input, K extends keyof Input>
+    extends AggregateBy<Input, K, Input[K] | undefined> {
+    constructor(inner: Operator<unknown, Input>, key: K) {
+        const min = (v1: Input[K] | undefined, v2: Input[K]) => {
+            if (v2 === undefined || v2 === null) {
```
espindola, line 403: Can v2 be undefined?

api/src/chisel.ts
```diff
@@ -347,6 +347,87 @@ class SortBy<T> extends Operator<T, T> {
...
+        const max = (v1: Input[K] | undefined, v2: Input[K]) => {
+            if (v2 === undefined || v2 === null) {
+                return v1;
+            }
+            if (v1 === undefined) {
+                return v2;
+            }
+            return v1 > v2 ? v1 : v2;
+        };
+        super(inner, key, undefined, max);
```
espindola, line 427: Does fold always start with undefined? What other aggregations do you have in mind? Do they all have the property that agg([x]) == x? Could you use that to simplify?

espindola (approved):

PR #1340, opened 2022-05-31
Output relative paths for local imports

espindola:
In the common case of endpoints/foo.ts importing models/bar.ts, the
output of compile_ts_code would contain "endpoint/foo.ts", but
"file:///path/to/models/bar.ts". That is inconvenient since the
absolute paths will be different in the server.

With this patch we output "endpoint/foo.ts" and "models/bar.ts", which
is the format we want when sending multiple modules to the server.

This also fixes the output path of a top level JS file. A testcase is
included.

### DIFF
--- a/tsc_compile/src/lib.rs
+++ b/tsc_compile/src/lib.rs
@@ -302,6 +302,7 @@ impl Compiler {
         file_name: &str,
         opts: CompileOptions<'_>,
     ) -> Result<HashMap<String, String>> {
+        let dir_path = Path::new(file_name).parent().unwrap().to_path_buf();
         let url = format!("file://{}", abs(file_name));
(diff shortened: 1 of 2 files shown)


athoscouto (commented):
The code looks good.

I'm only worried about possible unforeseen consequences of mutating `map.written`.
Any problems that could come from that?

espindola:
> The code looks good.
> 
> I'm only worried about possible unforeseen consequences of mutating `map.written`. Any problems that could come from that?

I don't think so. Note that map comes from

```
let mut map = op_state.take::<DownloadMap>();
```
So it is not a reference. We see all uses of it.

athoscouto (approved):

PR #1345, opened 2022-05-31
crud.rs: Sort by id by default.

BearLemma:
Introduces sorting by id by default in CRUD.
Requested by @espindola . Prerequisite to #1281 .


espindola:
Fine by me as we can add a way to get an unpaged, unsorted results later if this turns out to be too expensive. I just wanted to make this decision explicit. What do others think?

espindola (approved):

dekimir:
> Fine by me as we can add a way to get an unpaged, unsorted results later if this turns out to be too expensive. I just wanted to make this decision explicit. What do others think?

For the record: we discussed this verbally and decided that the user can always write a custom endpoint that doesn't sort.  But ChiselEntity.crud() now forces paging, so it must always sort by something.

PR #1348, opened 2022-05-31
Simplify makeTransformedQueryIter

espindola:


api/src/chisel.ts
```diff
@@ -468,18 +468,14 @@ export class ChiselCursor<T> {
     private makeTransformedQueryIter(
         op: Operator<T>,
     ): AsyncIterable<T> | undefined {
-        if (op.type == OpType.BaseEntity) {
+        if (op.inner === undefined) {
```
athoscouto, line 471: Are we assuming that `op.inner === undefined` implies in `op.type === OpType.BaseEntity`?
Is it safe to do so?

athoscouto (approved):

PR #1349, opened 2022-05-31
Delete OpType

espindola:
We can use reflection to set the type property. We could also use
reflection to de-serialize and drop the type property completely, but
it is a lot easier to let serde do it for us.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -4,26 +4,17 @@
 /// <reference lib="dom" />
 /// <reference lib="dom.iterable" />
 
-enum OpType {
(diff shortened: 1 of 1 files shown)


penberg (commented):
Looks good to me, but will let @BearLemma review

BearLemma:
I generally don't like the usage of `constructor.name` as it seems to be a bit of a minefield, but in this case, the operator types are fully under our control and intended for our use only so I don't think there is any danger. Approving :relaxed:

BearLemma (approved):

PR #1352, opened 2022-06-01
Pass multiple files to compile_ts_code

espindola:
This speeds chisel apply in a directory with two endpoints by 1.12X
and with 3 by 1.24X.

This is also on the path to avoid compiling on the server.


cli/src/cmd/apply/deno.rs
```diff
@@ -8,21 +8,24 @@ use anyhow::{Context, Result};
...
     use_chiselc: bool,
 ) -> Result<Vec<EndPointCreationRequest>> {
     let mut endpoints_req = vec![];
+    let paths: Vec<_> = endpoints
+        .iter()
+        .map(|f| f.file_path.to_str().unwrap())
+        .collect();
+    let mut output = compile_endpoint(&paths)
+        .await
+        .with_context(|| "parsing endpoints")?;
```
BearLemma, line 24: .context()? or better error saying which endpoint failed to compile or something?

cli/src/cmd/apply/deno.rs
```diff
@@ -8,21 +8,24 @@ use anyhow::{Context, Result};
...
     entities: &[String],
     types_string: &str,
     use_chiselc: bool,
 ) -> Result<Vec<EndPointCreationRequest>> {
     let mut endpoints_req = vec![];
+    let paths: Vec<_> = endpoints
+        .iter()
+        .map(|f| f.file_path.to_str().unwrap())
+        .collect();
+    let mut output = compile_endpoint(&paths)
```
BearLemma, line 22: `compile_endpoint` -> `compile_endpoints`?

cli/src/cmd/apply/deno.rs
```diff
@@ -8,21 +8,24 @@ use anyhow::{Context, Result};
...
+pub(crate) async fn apply(
     endpoints: &[Endpoint],
     entities: &[String],
     types_string: &str,
     use_chiselc: bool,
 ) -> Result<Vec<EndPointCreationRequest>> {
     let mut endpoints_req = vec![];
+    let paths: Vec<_> = endpoints
+        .iter()
+        .map(|f| f.file_path.to_str().unwrap())
```
BearLemma, line 20: Maybe check that the conversion to string succeeded? It's a user provided so it can be anything really... If you denote the paths type to be `Result<Vec<_>>`, collect will do the transposition on its own.

tsc_compile/src/lib.rs
```diff
@@ -260,6 +260,15 @@ impl Resolver for ModuleResolver {
     }
 }
 
+fn find_relative<'a>(urls: &HashMap<Url, &'a str>, source: &Url) -> Option<(String, &'a str)> {
```
BearLemma, line 263: A blurb would be nice explaining what's going on here

tsc_compile/src/lib.rs
```diff
@@ -300,12 +309,16 @@ impl Compiler {
...
-        file_name: &str,
+        file_names: &[&str],
         opts: CompileOptions<'_>,
     ) -> Result<HashMap<String, String>> {
-        let dir_path = Path::new(file_name).parent().unwrap().to_path_buf();
-        let url = format!("file://{}", abs(file_name));
-        let url = Url::parse(&url)?;
+        let mut url_to_name: HashMap<Url, &str> = HashMap::new();
+        let mut urls = Vec::new();
+        for n in file_names {
```
BearLemma, line 317: naming it `name` or `file_name` wouldn't hurt in this case.

tsc_compile/src/lib.rs
```diff
@@ -300,12 +309,16 @@ impl Compiler {
...
+            let u = Url::parse(&format!("file://{}", abs(n)))?;
```
BearLemma, line 318: also, the `u` is `url`...

tsc_compile/src/lib.rs
```diff
@@ -383,11 +399,12 @@ impl Compiler {
                 let prefix = without_extension(&k);
                 let is_dts = k.ends_with(".d.ts");
                 let source = prefix_map[prefix];
-                let source = if *source == url {
-                    file_name.to_string()
+                let source = if let Some(n) = url_to_name.get(source) {
```
BearLemma, line 402: It's name a bit further down, it wpuld be nice to be consistent.

cli/src/cmd/apply/deno.rs
espindola: From here it is hard to recover which endpoint had an error. TSC does print the exact file line and column, so I think we are not reducing the error quality. I will replace with context()

PR #1358, opened 2022-06-01
Don't map null to undefined in getSecret

espindola:
JSON has null, but not undefined, so it is odd to accept JSON but map
null to undefined.

If we really want to keep using undefined for "no secret was set", I can do it, but since that implementation is more complicated I wanted to check first.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -897,8 +897,8 @@ type JSONValue =
     | { [x: string]: JSONValue }
     | Array<JSONValue>;
 
-export function getSecret(key: string): JSONValue | undefined {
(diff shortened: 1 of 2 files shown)


athoscouto:
@penberg @dekimir any context on why Glauber mapped `null`s to `undefined`?

dekimir:
> @penberg @dekimir any context on why Glauber mapped `null`s to `undefined`?

I don't know.  It seems better to pass `null` through unchanged.  That way we distinguish between a missing value and a `null` value.

I checked the platform backend code, and I don't think this will break anything.

dekimir (approved):

espindola:
For the record, returning undefined is surprisingly hard in deno. I reported https://github.com/denoland/deno/issues/14779.

PR #1362, opened 2022-06-02
CRUD: Implement previous page

BearLemma:
This PR brings "prev_page" to CRUD response. Builds upon #1281 
I've divided it into two commits, but I'll give it a thought after lunch and see if I can split it further to improve readability of this PR.
Also some rests to be done:

- [x] add lit test coverage
- [x] add blurbs to some newly created functions

### DIFF
--- a/server/src/datastore/crud.rs
+++ b/server/src/datastore/crud.rs
@@ -852,41 +852,65 @@ mod tests {
         }
 
         // Test cursors
-        {
-            let mut page_url = url("page_size=1");
+        async fn run_cursor_test(
+            qe: &QueryEngine,
+            mut page_url: Url,
+            n_steps: usize,
+            page_size: usize,
+        ) -> Vec<String> {
+            let get_url = |raw: &serde_json::Value| Url::parse(raw.as_str().unwrap()).unwrap();
             let mut all_names = vec![];
-            for i in 0..5 {
+            for i in 0..n_steps {
                 let r = run_query("Person", page_url.clone(), qe).await.unwrap();
+
+                // Check backward cursors
+                if i == 0 || i == n_steps - 1 {
+                    // FIXME: The last page should always (except for no results at all) include
+                    // link for previous page.
+                    assert!(!r.contains_key("prev_page"));
+                } else {
+                    // Check that previous page returns the same results as
+                    // before using next_page.
+                    let prev_page = get_url(&r["prev_page"]);
+                    let r = run_query("Person", prev_page.clone(), qe).await.unwrap();
+                    let prev_names = collect_names(&r);
+                    assert_eq!(prev_names.len(), page_size);
+                    assert_eq!(prev_names, all_names[all_names.len() - page_size..]);
+
+                    // Check that a sequence page1 -> page["prev_page"] (page2) -> page2["next_page"] (page3)
+                    // is a cycle, i.e. page1 == page3.
+                    let loopback_page = get_url(&r["next_page"]);
+                    assert_eq!(page_url, loopback_page);
+
+                    if i == 1 {
+                        // We go from the second page to first and beyond to an empty page.
+                        let prev_page = get_url(&r["prev_page"]);
+                        let r = run_query("Person", prev_page.clone(), qe).await.unwrap();
+                        // FIXME: It should be possible to get back from an empty page.
+                        assert!(!r.contains_key("next_page"));
+                    }
+                }
+
                 let names = collect_names(&r);
                 all_names.extend(names.clone());
-                if i == 4 {
+                // Check forward cursors
+                if i == n_steps - 1 {
                     assert!(names.is_empty());
                     assert!(!r.contains_key("next_page"));
                 } else {
-                    assert_eq!(names.len(), 1);
-                    let next_page = r["next_page"].as_str().unwrap();
-                    page_url = Url::parse(next_page).unwrap();
+                    assert_eq!(names.len(), page_size);
+                    page_url = get_url(&r["next_page"]);
                 }
             }
+            all_names
+        }
+        {
+            let mut all_names = run_cursor_test(qe, url("page_size=1"), 5, 1).await;
             all_names.sort();
             assert_eq!(all_names, vec!["Alan", "Alex", "John", "Steve"]);
         }
         {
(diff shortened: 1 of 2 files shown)


server/src/datastore/crud.rs
```diff
@@ -47,58 +47,127 @@ fn run_query_impl(
...
+/// prev page url if there is a potential for retrieving elements that precede
+/// current resulting elements in query's sort.
+fn get_prev_page(
+    params: &QueryParams,
+    query: &Query,
+    host: &Option<String>,
+    results: &[JsonObject],
+) -> Result<Option<Url>> {
+    if results.is_empty() || query.cursor.is_none() {
+        return Ok(None);
```
espindola, line 113: Is this correct? If we are at the last page, results can be empty and we need a previous page link, no?

server/src/datastore/crud.rs
```diff
@@ -168,9 +237,16 @@ impl Query {
...
                     anyhow::ensure!(
                         q.cursor.is_none(),
-                        "only one occurrence of page_after is allowed."
+                        "only one occurrence of page_before/page_after is allowed."
                     );
-                    q.cursor = Cursor::from_string(&value)?.into();
+                    q.cursor = Cursor::from_string(&value, true)?.into();
+                }
+                "page_before" => {
+                    anyhow::ensure!(
```
espindola, line 245: Please don't duplicate the ensure. You can write this as
```
"page_after" | "page_before" => {
ensure...
 q.cursor = Cursor::from_string(&value, param_key== "page_after")?.into();
}
```

server/src/datastore/crud.rs
```diff
@@ -47,58 +47,127 @@ fn run_query_impl(
...
+/// Evaluates current query circumstances and potentially generates
+/// next page url if there is a potential for retrieving elements that succeed
+/// current resulting elements in query's sort.
+fn get_next_page(
+    params: &QueryParams,
+    query: &Query,
+    host: &Option<String>,
+    results: &[JsonObject],
+) -> Result<Option<Url>> {
+    if results.is_empty() {
```
espindola, line 90: Is this correct? With the current implementation I think you can get an empty page before all entries by following the previous page link, no?

server/src/datastore/crud.rs
```diff
@@ -47,58 +47,127 @@ fn run_query_impl(
...
     host: &Option<String>,
     query: &Query,
-    last_element: &JsonObject,
+    pivot_element: &JsonObject,
+    forward: bool,
 ) -> Result<Url> {
-    assert!(!query.sort.keys.is_empty());
+    // If cursor is available, we must use its sort keys as query's
+    // sort keys are reversed for backwards paging.
+    let sort_keys = if let Some(cursor) = &query.cursor {
```
espindola, line 140: Crazy idea that might simplify the code, no need to do it now: Drop sort from the Query structure and create a cursor pointing at the first page if we are given a sort parameter in the Url.

server/src/datastore/crud.rs
BearLemma: Good idea! Thanks.

server/src/datastore/crud.rs
BearLemma: I think that this could work. One small drawback is that I'll need to crate a Query flag indicating whether we are already cursor paging or not (so that I don't generate prev_page for the first page), but otherwise sounds good.

But let's deal with the corner cases first so that we have the full picture.

server/src/datastore/crud.rs
BearLemma: Yes, but the problem is that with no results I can't generate cursor as I have no pivot. 
There are ways to work around this (selecting page_size + 1, min/max pivot or just allowing the pivot to be omitted), but it's a really tricky matter.. So perhaps we can just ignore the problem for the time being?

server/src/datastore/crud.rs
espindola: I don't see why we would need another flag. If I open a query with sort, follow the next link and then the prev link, I should get the first page and no prev_page link. I know that there is a bug right now, but we should fix the bug, not special case it.

server/src/datastore/crud.rs
espindola: We can ignore it, but on the conditions that:

* There is a comment explaining why we currently return the wrong result if results.is_empty()
* There is a test showing that we return the wrong result and a FIXME about it
* We make an effort right after this is in to fix it. If this proves too hard, we upgrade the FIXME to an issue.

server/src/datastore/crud.rs
```diff
@@ -852,41 +852,68 @@ mod tests {
...
+            for i in 0..n_steps {
                 let r = run_query("Person", page_url.clone(), qe).await.unwrap();
+
+                // Check backward cursors
+                if i == 0 {
+                    assert!(!r.contains_key("prev_page"));
+                } else if i == n_steps - 1 {
+                    // FIXME: The last page should always (except for no results at all) include
+                    // link for previous page.
+                    // assert!(r.contains_key("prev_page"));
```
espindola, line 872: nit: Actually assert the wrong result, so that anyone fixing this sees the test failure and updates it too.

server/src/datastore/crud.rs
```diff
@@ -852,41 +852,68 @@ mod tests {
...
+                    // is a cycle, i.e. page1 == page3.
+                    let loopback_page = get_url(&r["next_page"]);
+                    assert_eq!(page_url, loopback_page);
+
+                    if i == 1 {
+                        // We go from the second page to first and beyond to an empty page.
+                        let prev_page = get_url(&r["prev_page"]);
+                        let _r = run_query("Person", prev_page.clone(), qe).await.unwrap();
+                        // FIXME: It should be possible to get back from an empty page.
+                        // assert!(r.contains_key("next_page"));
```
espindola, line 892: Same here.

espindola (commented):
OK by me with one last change

server/src/datastore/crud.rs
BearLemma: Fixed

server/src/datastore/crud.rs
BearLemma: Fixed

espindola (approved):

PR #1363, opened 2022-06-02
cli/apply: User feedback if `chiselc` cannot be found

penberg:
Add some user feedback if `chiselc` cannot be found so that they know
what to expect:

  Warning: no ChiselStrike compiler (`chiselc`) found. Some your queries will run slowly.
  Model defined: User
  End point defined: /dev/find
  End point defined: /dev/fixture
  End point defined: /dev/hello


penberg:
In draft mode because commit on top of #1290

cli/src/cmd/apply.rs
```diff
@@ -85,8 +85,12 @@ pub(crate) async fn apply<S: ToString>(
         .iter()
         .map(|type_req| type_req.name.clone())
         .collect();
-    let optimize = is_chiselc_available() && manifest.optimize == Optimize::Yes;
-    let auto_index = is_chiselc_available() && manifest.auto_index == AutoIndex::Yes;
+    let chiselc_available = is_chiselc_available();
+    if !chiselc_available {
+        println!("Warning: no ChiselStrike compiler (`chiselc`) found. Some your queries will run slowly.");
```
BearLemma, line 90: `Some of your queries might be slow`?

PR #1364, opened 2022-06-02
Simple benchmark application

penberg:
This adds a very simple application for benchmarking purposes. The main
benefit right now is the fixture setup using Faker.js, but we can keep
on improving this with more scenarios and more automation for comparing
performance.

### DIFF
--- a/perf/README.md
+++ b/perf/README.md
@@ -1,0 +1,22 @@
+# Performance Benchmarks
+
+## Getting Started
+
+Start ChiselStrike server:
+
+```console
+cd backend && ../../target/debug/chisel dev
+```
+
+Generate some test data:
+
+```console
+curl -X POST localhost:8080/dev/fixture
+```
+
+Benchmark:
+
+```console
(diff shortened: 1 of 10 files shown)


perf/backend/endpoints/hello.ts
```diff
@@ -0,0 +1,3 @@
+export default async function (req: Request): Promise<Response> {
```
dekimir, line 1: Is this one actually used?  If not, let's not check it in.

dekimir (changes requested):
Looks good, but drop hello.ts.

perf/README.md
```diff
@@ -0,0 +1,22 @@
...
+Start ChiselStrike server:
+
+```console
+cd backend && ../../target/debug/chisel dev
+```
+
+Generate some test data:
+
+```console
+for i in {1..1000}; do curl -X POST localhost:8080/dev/fixture; done
```
dekimir, line 14: Doesn't fixture already have its own loop?

penberg: @dekimir Yeah, fixed the README now.

dekimir (approved):

PR #1367, opened 2022-06-02
Return a JsonObject from getSecret

espindola:
Also use a simple comparison to check if we have a new one. We already
have a copy of the secrets in each worker, so using a hash was not
saving much memory. getSecret was also already decripting the secrets,
so avoiding a json parse was also not saving that much cpu.

With this all errors getting secrets are handled uniformly.

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -134,14 +134,12 @@ async fn run(state: SharedState, mut cmd: ExecutorChannel) -> Result<()> {
     init_deno(state.inspect_brk).await?;
 
     // Ensure we read the secrets before spawning an ApiService; secrets may dictate API authorization.
-    if let Ok(secrets) = get_secrets().await {
-        match serde_json::from_str(&secrets) {
-            Err(e) => {
-                warn!("Could not read secrets: {:?}", e);
-                update_secrets(Default::default()).await;
-            }
-            Ok(json) => update_secrets(json).await,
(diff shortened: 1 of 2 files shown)


server/src/server.rs
```diff
@@ -324,35 +313,28 @@ pub async fn run_shared_state(
...
-                        cmd.send(payload).await.unwrap();
-                    }
+                    last_secrets = secrets;
                 }
                 Err(x) => {
                     if !last_try_was_failure {
                         warn!("Could not read secrets: {:?}", x);
                     }
                     last_try_was_failure = true;
+                    last_secrets = JsonObject::default();
```
athoscouto, line 326: Why are you setting `last_secrets` when there is a failure?
Won't it effectively drop any secrets in case of a transient failure?

espindola: Yes, and that is expected by the secrets.deno test. I think it is reasonable, given that we log and reload when it is fixed. We can change the behavior, but we should probably do that in another PR.

BearLemma:
LGTM, if @athoscouto 's comment is resolved, I think we can merge this baby.

athoscouto (approved):

espindola:
Could you open an issue? Glauber coded this and the test, so I would
like to check with him why he went for this behavior.

Cheers,
Rafael

PR #1368, opened 2022-06-02
Don't flatten IdTree

espindola:
This simplifies processing it.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -607,17 +607,14 @@ export class ChiselEntity {
             name: this.constructor.name,
             value: this,
         }, requestContext);
-        type IdsJson = Map<string, IdsJson>;
+        type IdsJson = { id: string; children: { [index: string]: IdsJson } };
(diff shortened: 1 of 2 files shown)


api/src/chisel.ts
```diff
@@ -607,17 +607,14 @@ export class ChiselEntity {
             name: this.constructor.name,
             value: this,
         }, requestContext);
-        type IdsJson = Map<string, IdsJson>;
+        type IdsJson = { id: string; children: { [index: string]: IdsJson } };
```
athoscouto, line 610: :pick:: what about `type IdTree = { id: string; children: Map<string, IdTree> };`?

The name matches the Rust struct and the map makes it a little easier to read.

athoscouto (approved):

api/src/chisel.ts
espindola: I don't think Map is correct in here, it is not the same, for example:

```
const a = {};
const b = new Map();
a.entries // undefined
b.entries // Function
```

It would probably be correct to use `Record<string, IdTree>`. Do you think that is more idiomatic?

PR #1372, opened 2022-06-03
Add type safe opSync and opAsync

espindola:
I have fixed most type errors already. This fixes a null X undefined
error and adds a FIXME for the last one I know of.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -4,6 +4,14 @@
 /// <reference lib="dom" />
 /// <reference lib="dom.iterable" />
 
+function opSync(opName: string, a?: unknown, b?: unknown): unknown {
+    return Deno.core.opSync(opName, a, b);
+}
+
+function opAsync(opName: string, a?: unknown, b?: unknown): Promise<unknown> {
+    return Deno.core.opAsync(opName, a, b);
+}
+
 /**
  * Base class for various Operators applicable on `ChiselCursor`.
  */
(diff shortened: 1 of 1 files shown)


api/src/chisel.ts
```diff
@@ -481,19 +489,19 @@ export class ChiselCursor<T> {
...
                     while (true) {
-                        const properties = await Deno.core.opAsync(
+                        const properties = await opAsync(
                             "op_chisel_query_next",
                             rid,
-                        );
+                        ) as T | null; // FIXME: This is wrong, we can get less than T with ColumnsSelect.
 
-                        if (properties == undefined) {
+                        if (properties === null) {
```
BearLemma, line 504: Why is is this changed to null?

dekimir:
So `(a?: unknown, b?: unknown): unknown` is type safe? ;)

espindola:
It is. You have to explicitly convert unknown to another type before
using, so we can just search for type asserts to see where errors might
be.

The existing Deno.core.opSync/opAsync use any instead of unknown.

Cheers,
Rafael

BearLemma:
Interesting, I wrote the exact same thing via email, but it didn't work for me for some reason 🤔  @espindola you replied via email, right? And did you click reply all or just reply?

api/src/chisel.ts
espindola: Because it is more precise. We were returning null already, but the comparison would pass because of the == instead of ===

BearLemma:
I think I did that as well :thinking:. Let's see if it works this time.

BearLemma (approved):

espindola:
Because that is what we get from rust. The old code worked because it
was using == instead of ===.

Cheers,
Rafael

PR #1374, opened 2022-06-03
Fix model dependency bug

MarinPostma:
This PR solves a bug related to dependency in models:

Let's say I have:

```typescript
// in `models/A.ts`
import { ChiselEntity } from "@chiselstrike/api"
import { B } from "./B.ts"

export class A extends ChiselEntity {
	b: B;
}
```

```typescript
// in `models/B.ts`
import { ChiselEntity } from "@chiselstrike/api"

export class B extends ChiselEntity {
	field: string;
}
```

you would get the error ``Error: field type `B` is neither a built-in nor a custom type``. This error doesn't occur if you rename `A.js` to `B.js`.

This is fixed here by performing a topological sort of the `AddTypeRequest` so that further checks are performed in an order that satisfies the dependencies between the models.

This also allows us to detect cyclic dependencies between models.


penberg:
@MarinPostma There's a conflict on `Cargo.lock` for some reason. Please rebase against `main`?

penberg:
@espindola looks like perhaps something for you to review?

MarinPostma:
I struggled rebasing and ended up creating a new branch on this repository instead of my fork. This pr is moved to #1443

PR #1379, opened 2022-06-06
CRUD: paging on empty pages

BearLemma:
This PR does two things.
1) it provides previous/next paging for all pages (even the first one). The rationale is twofold for this. a) it's simpler and b) there actually always can be some results anywhere as new instances can be inserted anytime while paging.
2) It uses the concept of cursor reversal to be able to come back from empty pages (during reversal of a cursor, the direction is reversed and it's made inclusive so that it includes the pivot (last/first element) of the page we came from). 

TODO:

- [x] Add a bit more test coverage for the new behavior.

### DIFF
--- a/server/src/datastore/crud.rs
+++ b/server/src/datastore/crud.rs
@@ -87,12 +87,7 @@ fn get_next_page(
     host: &Option<String>,
     results: &[JsonObject],
 ) -> Result<Option<Url>> {
-    if !results.is_empty() {
-        let last_element = results.last().unwrap();
-        let url = make_page_url(&params.url, host, query, last_element, true)?;
-        return Ok(Some(url));
-    }
-    Ok(None)
+    get_page(params, query, host, results, true)
 }
 
 /// Evaluates current query circumstances and potentially generates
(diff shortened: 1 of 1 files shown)


server/src/datastore/crud.rs
```diff
@@ -87,12 +87,7 @@ fn get_next_page(
     host: &Option<String>,
     results: &[JsonObject],
 ) -> Result<Option<Url>> {
-    if results.is_empty() {
-        return Ok(None);
-    }
-
-    let has_bwd_cursor = query.cursor.as_ref().map(|c| !c.forward).unwrap_or(false);
-    if query.page_size == results.len() as u64 || has_bwd_cursor {
+    if !results.is_empty() {
```
espindola: The commit message says "Always provide next/prev page", but the patch doesn't do that for empty pages, which one is wrong?

server/src/datastore/crud.rs
```diff
@@ -109,12 +104,7 @@ fn get_prev_page(
     host: &Option<String>,
     results: &[JsonObject],
 ) -> Result<Option<Url>> {
-    if results.is_empty() || query.cursor.is_none() {
-        return Ok(None);
-    }
-
-    let has_fwd_cursor = query.cursor.as_ref().map(|c| c.forward).unwrap_or(false);
-    if query.page_size == results.len() as u64 || has_fwd_cursor {
+    if !results.is_empty() {
```
espindola, line 112: Same here.

server/src/datastore/crud.rs
espindola: Can you expand the commit message a bit on the why of the patch?

BearLemma: Added more info to the commit message.

espindola (approved):

PR #1380, opened 2022-06-07
Refactor secret reading to a helper function

espindola:
It is an open issue whether we want to treat failures in both
locations the same, but while we are, may as well do so from a single
function.

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -130,17 +130,30 @@ pub(crate) async fn add_endpoint<S: AsRef<str>>(
     Ok(())
 }
 
+async fn read_secrets() -> JsonObject {
+    static LAST_TRY_WAS_FAILURE: Mutex<bool> = Mutex::new(false);
+    let secrets = get_secrets().await;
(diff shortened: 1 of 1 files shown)


server/src/server.rs
```diff
@@ -130,17 +130,30 @@ pub(crate) async fn add_endpoint<S: AsRef<str>>(
...
+    let secrets = get_secrets().await;
+    let mut was_failure = LAST_TRY_WAS_FAILURE.lock().await;
+    match secrets {
+        Ok(secrets) => {
+            *was_failure = false;
+            secrets
+        }
+        Err(e) => {
+            if !*was_failure {
+                warn!("Could not read secrets: {:?}", e);
```
BearLemma, line 144: Why not war at each error?

espindola: I assume it is to avoid having too many warning on the log, but note this was already the case.

espindola: The first case is only used once at startup.

BearLemma (approved):

PR #1383, opened 2022-06-07
chiselc: Add test for transforming filter with side-effects

penberg:
Test that transforming a filter with side-effects preseves the lambda.

### DIFF
--- a/chiselc/tests/lit/transform-filter-with-side-effects.lit
+++ b/chiselc/tests/lit/transform-filter-with-side-effects.lit
@@ -1,0 +1,10 @@
+// Test transforming a filter with side-effects.
+// RUN: @chiselc @file -e Person
+
+await Person.cursor().filter(p => {
+    return p.age > 20 && await fetch('http://example.com/movies.json');
+});
+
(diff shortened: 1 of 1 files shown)


chiselc/tests/lit/transform-filter-with-side-effects.lit
```diff
@@ -0,0 +1,10 @@
+// Test transforming a filter with side-effects.
+// RUN: @chiselc @file -e Person
+
+await Person.cursor().filter(p => {
+    return p.age > 20 && await fetch('http://example.com/movies.json');
+});
+
+// CHECK: await Person.cursor().filter((p)=>{
+// CHECK:     return p.age > 20 && await fetch('http://example.com/movies.json');
```
BearLemma, line 9: It would be great if the compiler was able to first prefilter by the age filter and then postfilter with the dynamic check. The problem is that this is only possible for And operator :(

penberg: @BearLemma Yes, that's a great feature and I hope to work on it. But this pull request is about adding a test case.

BearLemma (approved):

PR #1387, opened 2022-06-08
Consolidate serde versions

penberg:


espindola (approved):

espindola:
Idea for a followup: Have a common deps crate that just reexports some dependencies, so that we don't have versions written in multiple places.

PR #1391, opened 2022-06-08
change perf benchmark

glommer:
* adds an age field, so we can demonstrate range queries as well
* benchmarks from the release build
* make the fixture a bit smarter, and add the elements to be found there
  for simplicity

### DIFF
--- a/perf/README.md
+++ b/perf/README.md
@@ -2,10 +2,15 @@
 
 ## Getting Started
 
+Build the release version for better performance
+
+```console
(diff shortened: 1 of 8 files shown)


penberg (approved):

glommer:
failed linter - I didn't even know the linter would reach that directory, but good that it does!
I will update the example anyway with some other minor changes, and resend.

penberg (approved):

Issue #1204, opened 2022-04-27
Add support for cursor-based pagination


penberg:
Cursor-based paging avoids the issue of duplicate results and supports infinite scrolling use case. Furthermore, Prisma recently added support for it as well: https://www.prisma.io/docs/concepts/components/prisma-client/pagination

Issue #610, opened 2022-01-15
Default paging limit


penberg:
Kyle Mathews suggested to add a default limit to paging, which is big enough for small collections because the simpler API facilitates adoption as people play around with ChiselStrike.

PR #1400, opened 2022-06-09
Update our docs to cover paging.

BearLemma:
This PR adds coverage of paging to our docs and modifies all examples to reflect the fact that CRUD response has been changed from `[...]` to `{"results": [...]}`.

closes https://github.com/chiselstrike/chiselstrike/issues/1204
closes https://github.com/chiselstrike/chiselstrike/issues/610

### DIFF
--- a/docusaurus/docs/Intro/endpoints.md
+++ b/docusaurus/docs/Intro/endpoints.md
@@ -6,8 +6,8 @@ In this section we'll show how to move beyond simple CRUD requests, as shown in 
 
 <!-- FIXME: move into extra chapter? -->
 
-CRUD generation is customizable; more detail and syntax around this and also security policy is coming soon but 
-here is a lower-level example that forbids DELETE, POST, and PUT while wrapping the GET result 
+CRUD generation is customizable; more detail and syntax around this and also security policy is coming soon but
+here is a lower-level example that forbids DELETE, POST, and PUT while wrapping the GET result
 with either `{"data": VALUE}` or `{"error": "message"}` depending on the result.
 
 <!-- FIXME: replace with class based alternates once available -->
@@ -27,7 +27,7 @@ export default crud(
         createResponses: {
             GET: (body: unknown, status: number) => {
                 if (status < 400) {
-                    return responseFromJson({ data: body }, status);
+                    return responseFromJson({ data: body["results"] }, status);
                 }
                 return responseFromJson({ error: body }, status);
             },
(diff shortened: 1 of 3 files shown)


docusaurus/docs/Intro/endpoints.md
```diff
@@ -126,32 +126,119 @@ curl -s localhost:8080/dev/comments
...
+
+### Why cursor-based paging?
+
+Compared to the classical offset-based paging, cursor paging has two main benefits.
+
+First big advantage is that cursor-based paging is **stable**. This means that if you do insertions
+resp. deletions while transitioning between pages, you won't miss entries resp. won't get duplicates.
+Those problems can be very annoying to deal with.
+
+Second advantage is **efficiency**. When paging using offsets, usually the whole query needs to be
```
espindola, line 231: Is this actually the case? In one case the DB is told to skip the first 100 elements, in the other it is told to skip elements with, for example, name < "bob". Both should be slow without indexes, and both should be fast with an index matching the sort we are using, no?

In any case, this paragraph seems more than what we need. We want to only offer cursor based paging, so I think just the high level description is OK.

BearLemma: Not quite. If you use any additional filtering, the offset case needs to iterate through all previous elements because it doesn't know which will match the filter and hence count into the offset. With cursor you simply jump to pivot and iterate-filter from there.

docusaurus/docs/Intro/endpoints.md
```diff
@@ -126,32 +126,119 @@ curl -s localhost:8080/dev/comments
...
+leverages the user-specified sort (primary key sorting is used if no sort is specified) and uses the
+elements as pivots. This way we naturally take advantage of indexes making everything much faster.
+
+### HOST header
+
+To construct the next/prev page links, we need to know what host and possibly port to use. It's not
+trivial to retrieve it automatically due to proxies etc., hence we utilize the `HOST` HTTP header
+from your request. For example `curl` sets it automatically and many HTTP libraries do the same.
+
+If `HOST` header is not specified, we will try to guess it, but it's highly recomended that it's
```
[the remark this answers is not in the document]
BearLemma: It hasn't changed so it's a very bad guess.

docusaurus/docs/Intro/endpoints.md
espindola: I see. In which case you can expand the paragraph or drop it. I would say it is probably a bit better to drop, but up to you.

docusaurus/docs/Intro/endpoints.md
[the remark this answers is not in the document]
espindola: :-)

docusaurus/docs/Intro/endpoints.md
BearLemma: Modified the paragraph. I would like to keep it as it's trying to demonstrate why cursor-based pagination is awesome because I've put a lot of effort into so I want to sell high :-D

espindola (approved):

PR #1403, opened 2022-06-09
Drop name from Endpoint

espindola:
With this the server is responsible for converting
endpoints/foo/bar.ts to foo/bar. This is in preparation to sending
more sources to the server.


utils/src/lib.rs
```diff
@@ -1,5 +1,12 @@
 use anyhow::{ensure, Result};
 use reqwest::{Response, Url};
+use std::path::Path;
+
+// Drop the extension (.ts/.js) from a path
+pub fn without_extension(p: &str) -> &str {
+    let ext_size = Path::new(p).extension().map(|s| s.len() + 1).unwrap_or(0);
```
BearLemma, line 7: Maybe you can use https://doc.rust-lang.org/std/path/struct.Path.html#method.with_extension so just to
`p.with_extension("")` ?

cli/src/project.rs
```diff
@@ -99,28 +99,22 @@ impl Manifest {
...
-                ret.push(Endpoint { file_path, name });
+                ret.push(Endpoint { file_path });
             }
         }
         ret.sort_unstable();
+        // Check for duplicated endpoints now since otherwise TSC
+        // reports the issue and we can produce a better diagnostic
+        // than TSC.
+        let i = ret.iter();
+        for (a, b) in i.clone().zip(i.skip(1)) {
```
BearLemma, line 111: Why do you need to clone? 🤔

utils/src/lib.rs
BearLemma: Or probably even bettter - https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem

cli/src/project.rs
dekimir: Maybe because `i` is consumed inside `zip(...)`?

BearLemma: Zip can't zip references? That should be possible, shouldn't it?

espindola: Note that it clones the iterator only. As far as I can tell that is needed: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.zip

PR #1405, opened 2022-06-10
chiselc: Filter predicate function splitting

penberg:
This patch adds support for splitting a filter predicate funcition into
a pure (no side-effects) and impure (potentially has side-effects) parts
so that we can transform the pure part into a runtime-optimized query.

For example, the following query:

```typescript
Person.cursor().filter(person => person.age > 40 && fetch("foobar"))
```

is transformed into:

```typescript
Person.cursor().__filter((person)=>person.age > 40
, {
    exprType: "Binary",
    left: {
        exprType: "Property",
        object: {
            exprType: "Parameter",
            position: 0
        },
        property: "age"
    },
    op: "Gt",
    right: {
        exprType: "Literal",
        value: 40
    }
}).filter((person)=>fetch("foobar")
);
```


chiselc/tests/lit/filter_splitting/filter-with-side-effects-unoptimizable.lit
```diff
@@ -0,0 +1,7 @@
+// Test splitting a filter with side-effects that cannot be split.
+// RUN: @chiselc @file -e Person
+
+Person.cursor().filter(person => person.age > 40 || fetch("foobar"))
+
+// CHECK: Person.cursor().filter((person)=>person.age > 40 || fetch("foobar")
+// CHECK: );
```
BearLemma, line 7: Newlines.. Perhaps check the "add newlines at the end of file" in vscode?

chiselc/tests/lit/filter_splitting/filter-with-side-effects-return.lit
```diff
@@ -0,0 +1,25 @@
+// Test splitting a filter with side-effects.
+// RUN: @chiselc @file -e Person
+
+Person.cursor().filter(person => { return person.age > 40 && fetch("foobar"); })
```
BearLemma, line 4: Perhaps test even the reversed order?

chiselc/src/query.rs
```diff
@@ -82,6 +82,10 @@ pub struct Filter {
     pub function: String,
     /// The original call expression of the filter.
     pub call_expr: CallExpr,
+    /// The pure (no side-effects) part of the filter expression AST.
+    pub pure: Box<swc_ecmascript::ast::Expr>,
```
BearLemma, line 86: What if there is no pure part? The filter operator is not generated at all, correct?

chiselc/tests/lit/filter_splitting/filter-with-side-effects-return.lit
penberg: done

chiselc/tests/lit/filter_splitting/filter-with-side-effects.lit
```diff
@@ -0,0 +1,23 @@
...
+// CHECK:         },
+// CHECK:         property: "age"
+// CHECK:     },
+// CHECK:     op: "Gt",
+// CHECK:     right: {
+// CHECK:         exprType: "Literal",
+// CHECK:         value: 40
+// CHECK:     }
+// CHECK: }).filter((person)=>fetch("foobar")
+// CHECK: );
```
BearLemma, line 23: Also no newline here.

chiselc/src/transforms/filter_splitting.rs
```diff
@@ -0,0 +1,108 @@
...
+        Expr::Member(member_expr) => is_pure_member_expr(member_expr),
+        Expr::Lit(_) => true,
+        _ => false,
+    }
+}
+
+/// Checks if the `member_expr` is pure (has no side-effects).
+fn is_pure_member_expr(member_expr: &MemberExpr) -> bool {
+    matches!(
+        (&*member_expr.obj, &member_expr.prop),
```
BearLemma, line 67: maybe `matches!() && matches!()` would be slightly more readable, but don't have a strong preference.

chiselc/tests/lit/filter_splitting/filter-with-side-effects-unoptimizable.lit
BearLemma: Indeed. The same applies for trailing whitespaces. I think we could just add a very annoying linter that checks lit files :thinking:

chiselc/src/transforms/filter_splitting.rs
penberg: done

chiselc/src/transforms/filter_splitting.rs
```diff
@@ -0,0 +1,135 @@
...
+/// Checks if the `expr` is pure (has no side-effects).
+///
+/// Note: we determine an expression to be pure conservatively. That is, we
+/// assume a function call, for example, to always have side-effects.
+fn is_pure_expr(expr: &Expr) -> bool {
+    match expr {
+        Expr::Bin(bin_expr) => is_pure_expr(&bin_expr.left) && is_pure_expr(&bin_expr.right),
+        Expr::Unary(unary_expr) => is_pure_expr(&unary_expr.arg),
+        Expr::Member(member_expr) => is_pure_member_expr(member_expr),
+        Expr::Lit(_) => true,
```
BearLemma, line 59: As discussed on Slack, we currently only allow for literal expressions to be pure, but it should be possible to allow variables to be used as well. Not necessary to fix it in this PR though.

BearLemma (approved):

chiselc/src/query.rs
```diff
@@ -82,6 +82,10 @@ pub struct Filter {
...
+    /// The impure (possible side-effects) part of the filter expression AST.
+    pub impure: Option<Box<swc_ecmascript::ast::Expr>>,
```
dekimir, line 88: "Part" is rather unspecified, unless we assume that `call_expr` is normalized (eg, conjunctive normal form).  A general expression can have many pure subexpressions connected by a complex operator graph.  Reading this blurb, I can't tell what precisely these parts are supposed to be.

ETA: #1416 is just one specific example of this, but I bet there's others.

chiselc/src/transforms/filter_splitting.rs
```diff
@@ -0,0 +1,135 @@
...
+        Expr::Bin(BinExpr {
+            op: BinaryOp::LogicalAnd,
+            left,
+            right,
+            span: _,
+        }) => {
+            if is_pure_expr(left) && !is_pure_expr(right) {
+                (*left.to_owned(), Some(*right.to_owned()))
+            } else {
+                (expr.to_owned(), None)
```
dekimir, line 43: This makes no sense to me.  How do we know `expr` is pure here??  Or is the first element the impure part?  Then the blurb is wrong and line 41 is wrong.

dekimir (commented):
Why are we focusing on purity?  Anything we can translate to SQL is pure by definition.  And anything we cannot (for whatever reason) must be evaluated in Rust, however pure it may be.

penberg:
@dekimir I focused on purity because I found it easy to reason about, but you are absolutely right that it's better to make this conditional to "can I transform this into a query expression" (`chiselc` doesn't know about SQL).

Btw, this is also why we have missed optimization bugs like this: https://github.com/chiselstrike/chiselstrike/issues/1412

chiselc/src/transforms/filter_splitting.rs
penberg: It does not make sense to you because you are absolutely right: we don't know if it's pure. But the way the current code is structured is that it doesn't matter because immediately after `split_expr` we do `convert_filter_expr` which will figure out if an expression cannot be transformed to a query expression, and just leave the code as-is.

But yeah this is confusing, and changing the code from "is this pure" to "is this transformable" will clear up the confusion.

PR #1407, opened 2022-06-13
tests: Fix CHECK in duplicated.deno

penberg:
Filesystem does not guarantee an ordering on the files so use a regexp
in the CHECK to deal with filenames appearing in different order.

### DIFF
--- a/cli/tests/lit/duplicated.deno
+++ b/cli/tests/lit/duplicated.deno
@@ -17,4 +17,4 @@ EOF
 cd "$TEMPDIR"
 $CHISEL apply 2>&1 || true
 
-# CHECK: Error: Cannot add both endpoints/foo.js endpoints/foo.ts as routes. ChiselStrike uses filesystem-based routing, so we don't know what to do. Sorry!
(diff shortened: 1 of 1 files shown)


espindola:
Should we sort instead? I somewhat like making chisel as predictable as possible.

penberg:
> Should we sort instead? I somewhat like making chisel as predictable as possible.

While I fully agree with the principle, I don't think sorting is justified here. The order of processing endpoints makes no difference to the developer, it's only a problem for `lit` tests. And in fact, you probably don't want to enforce any ordering of endpoint processing, because that prevents you from parallelization.

espindola:
> > Should we sort instead? I somewhat like making chisel as predictable as possible.
> 
> While I fully agree with the principle, I don't think sorting is justified here. The order of processing endpoints makes no difference to the developer, it's only a problem for `lit` tests. And in fact, you probably don't want to enforce any ordering of endpoint processing, because that prevents you from parallelization.

The processing itself is parallel. This error is from a quick pre check that just avoids a more confusing error from TSC. Give me a sec, I will send you an alternative PR to see what you think.

espindola:
https://github.com/chiselstrike/chiselstrike/pull/1403 should fix this

penberg:
Fixed by #1403, closing.

PR #1410, opened 2022-06-13
Make inner typing of our Cursor API more restrictive

BearLemma:


BearLemma:

api/src/chisel.ts
```diff
@@ -308,16 +314,16 @@ export class ChiselCursor<T> {
...
     /**
      * Restricts this cursor to contain just the objects that match the `Partial`
      * object `restrictions`.
      */
-    filter(restrictions: Partial<T>): ChiselCursor<T>;
+    filter(restrictions: Partial<ValueType>): ChiselCursor<BaseType, ValueType>;
 
     // Common implementation for filter overloads.
-    filter(arg1: ((arg: T) => boolean) | Partial<T>): ChiselCursor<T> {
+    filter(arg1: ((arg: ValueType) => boolean) | Partial<ValueType>) {
```
espindola, line 326: I would keep the explicit return type.

api/src/chisel.ts
```diff
@@ -450,12 +463,22 @@ export class ChiselCursor<T> {
...
             iter = this.runChiselQuery(this.inner);
         }
-        return iter[Symbol.asyncIterator]();
+        const typedIter = {
+            [Symbol.asyncIterator]: async function* () {
+                if (iter === undefined) {
+                    return;
+                }
+                for await (const e of iter) {
+                    yield e as ValueType;
```
espindola, line 477: This type assertion is now wrong, no?

api/src/chisel.ts
```diff
@@ -273,31 +279,31 @@ class SortBy<T> extends Operator<T> {
 }
 
 /** ChiselCursor is a lazy iterator that will be used by ChiselStrike to construct an optimized query. */
-export class ChiselCursor<T> {
+export class ChiselCursor<BaseType, ValueType, Input = ValueType> {
```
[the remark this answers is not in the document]
espindola: Sure, it is just the name that is odd. It makes me think of curiously recurring template pattern. I suggested ObjType since there is the "POD" `{a: number}` and there is the "complete" object that also has methods like save.

PR #1414, opened 2022-06-13
Don't reread the initial state

espindola:
With this we only read the endpoints, models and policies once during
startup.


penberg:
@espindola But won't this break `chisel dev` when I add a new endpoint, for example?

espindola:
> @espindola But won't this break `chisel dev` when I add a new endpoint, for example?

No. Chisel dev effectively issues 'chisel apply' on demand. The two reads were happening early on startup.

penberg (approved):

PR #1421, opened 2022-06-14
chiselc: Fix find{One,Many} filter splitting

penberg:
The `find{Many,One}` methods don't return a `ChiselCursor`, which is why
we can't chain an additional `filter`. Instead, change the internal
functions to take a third parameter -- postPredicate -- that contains
the part of a predicate that may have side-effects.

Bug spotted by Glauber.


penberg:
Depends on https://github.com/chiselstrike/chiselstrike/pull/1418

BearLemma (approved):

PR #1424, opened 2022-06-14
Change the code_map urls we use

espindola:
Currently an endpoint /api_ver/foo gets its source stored in a
hashtable keyed with the url
file:///api_ver/foo.js?ver=<file_version>.

Since that same hash table will have to store modules that are not
endpoints, this PR changes the url to
file:///api_ver/endpoints/foo.js?ver=<file_version>.

Which can also be used to store models or other files imported by the
endpoints.

There is a small hack in that we still store in the database using the
old format. This will be changed in a followup PR since that part
requires a migration.


espindola:
Putting this as draft as it has to go after #1423

penberg (approved):

penberg (approved):

Issue #1201, opened 2022-04-23
Can't use `labels` in multiple `models` files


dekimir:
When two different `models` files both import `labels` and one uses it, then `chisel apply` fails with this message:
```
Error: parsing endpoint /dev/po

Caused by:
    TypeError: Uncaught SyntaxError: Identifier 'labels' has already been declared
        at file:///dev/po.js?ver=2:5:9
        at async importEndpointImpl (file:///worker.js:39:17)
        at async handleMsg (file:///worker.js:13:17)
```

This is the minimal example that reproduces it:
```bash
$ cat models/person.ts 
import { labels } from "@chiselstrike/api";
export class Person {
  @labels("pii") last_name: string = "";
}
$ cat models/post.ts 
import { labels } from '@chiselstrike/api';
$ cat endpoints/po.ts 
export default async function (req: Request) {}
```

espindola:
mergify got this wrong

PR #1440, opened 2022-06-16
Restart on apply

espindola:
This solves the leak caused by never unloading old modules. It also avoids the somewhat surprising behavior that if `http://example.com/some/module.ts` is updated, we don't reload it until a restart. Last but not least, this opens the way to sending models as plain JS modules, which will fix https://github.com/chiselstrike/chiselstrike/issues/1201.

### DIFF
--- a/api/src/worker.ts
+++ b/api/src/worker.ts
@@ -75,7 +75,7 @@ function importEndpoints(endpoints: [Endpoint]) {
 
 async function importEndpointsImpl(endpoints: [Endpoint]) {
     for (const endpoint of endpoints) {
-        const { path, apiVersion, version } = endpoint;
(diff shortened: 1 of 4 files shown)


espindola:
> @espindola have you tested how it's going to affect the apply times? I'm asking because it's not that big of a deal for explicit invocation of apply, but it's going to affect `chisel dev` as well, where it might be inconvenient if it takes considerable time

TSC is so slow that it seems to dominate the `chisel apply` time. When running `chisel apply` with only a hello world endpoint,  on master perf reports `1.13245 +- 0.00536 seconds` and with this branch `1.19481 +- 0.00799 seconds`, so about 1.06 times slower.

PR #1443, opened 2022-06-17
Fix model dependency bug

MarinPostma:
This PR solves a bug related to dependency in models:

Let's say I have:

```typescript
// in `models/A.ts`
import { ChiselEntity } from "@chiselstrike/api"
import { B } from "./B.ts"

export class A extends ChiselEntity {
	b: B;
}
```

```typescript
// in `models/B.ts`
import { ChiselEntity } from "@chiselstrike/api"

export class B extends ChiselEntity {
	field: string;
}
```

you would get the error ``Error: field type `B` is neither a built-in nor a custom type``. This error doesn't occur if you rename `A.js` to `B.js`.

This is fixed here by performing a topological sort of the `AddTypeRequest` so that further checks are performed in an order that satisfies the dependencies between the models.

This also allows us to detect cyclic dependencies between models.


MarinPostma:
recheck

cli/tests/lit/cyclic-models.deno
```diff
@@ -0,0 +1,29 @@
+# SPDX-FileCopyrightText: © 2021 ChiselStrike <info@chiselstrike.com>
```
dekimir, line 1: 2022

cli/tests/lit/out-of-order-models.deno
```diff
@@ -0,0 +1,40 @@
+# SPDX-FileCopyrightText: © 2021 ChiselStrike <info@chiselstrike.com>
```
dekimir, line 1: 2022

server/src/rpc.rs
```diff
@@ -519,6 +521,43 @@ or
     }
 }
 
+fn sort_custom_types(
```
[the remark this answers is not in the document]
dekimir: I'm not sure I like either option.  Would it be possible to neither modify nor consume the request (or its parts)?  That way the caller can continue working with the request intact (eg, for logging, diagnostics, or additional processing in the future).

dekimir (approved):
A few nitpicks can be fixed in follow-ups.  The sorting function API can be argued here before we agree on whether and how to change it.

MarinPostma (commented):
The types were already consumed anyway on l326, by the `for`, but we could clone the types and work on that also, would that work for you?

server/src/rpc.rs
```diff
@@ -519,6 +521,43 @@ or
...
+        for field in &ty.field_defs {
+            if ts.lookup_builtin_type(&field.field_type).is_err() {
+                graph.add_node(field.field_type.as_str());
+                graph.add_edge(field.field_type.as_str(), ty.name.as_str(), ());
+            }
+        }
+    }
+
+    let order = petgraph::algo::toposort(&graph, None)
+        .map_err(|_| anyhow::anyhow!("cycle detected in models"))?
```
dekimir, line 543: What if one of the edges is an optional field? We could theoretically handle a cycle in that case.

dekimir:
> The types were already consumed anyway on l326, by the `for`, but we could clone the types and work on that also, would that work for you?

Yes. This operation isn't in a tight loop, it can afford the clone.

server/src/rpc.rs
MarinPostma: _Theoretically_, I think we _could_ handle cycles in the general case, since these kinds of cycles are entirely OK to represent in the relational model.

However, some rework on the current implementation would need to be done at the type level. This is because the `Type` type is recursive: it has `fields` that contain a `Type` themselves, and we would end up with infinite type recursion. 

I don't understand enough the implication of changing that right now, this probably needs further discussion.

espindola: > However, some rework on the current implementation would need to be done at the type level. This is because the `Type` type is recursive: it has `fields` that contain a `Type` themselves, and we would end up with infinite type recursion.
> 
> I don't understand enough the implication of changing that right now, this probably needs further discussion.
 
Sounds like a new issue is called for :-)

PR #1449, opened 2022-06-20
Fix issue generating lockfile

MarinPostma:
This PR fixes an issue that occur when generating the `Cargo.lock` file. This is caused by `swc_ecmascript_codegen_macro` not being compliant with semver, and breaking the build when bumping the patch number (which cargo expect to be able to do safely).

This is probelmatic because it prevents us to bump our dependency correctly to the next semver compatible version (`cargo update`), and to regenerate a `Cargo.lock` file, which comes handy when fixing merge conflict on it, which should be pretty common.

Since we can now safely bump the dependences, I did so in another commit.


MarinPostma:
I am trying to fix the failing test. `cargo-udeps` seems to fail to compile the project due to, conflicting dependencies. This looks like a resolver issue in cargo udeps. I'm checking that out.

chiselc/src/main.rs
```diff
@@ -15,6 +15,13 @@ use std::io::{self, Read, Write};
 use std::path::PathBuf;
 use structopt::StructOpt;
 
+// We import that here only to make cargo udeps happy. We need to do that because swc released a
+// breaking version of `swc-ecma-codegen-macros`.
+mod _import_emmiter {
+    #[allow(unused)]
```
espindola, line 21: Do we need this? `cargo test -p cli --test linters unused_dependencies` passes for me without it.

espindola (commented):
LGTM with the 3rd patch dropped.

chiselc/src/main.rs
MarinPostma: you're right, it's not needed anymore

espindola (approved):

PR #1455, opened 2022-06-21
Add a library directory to projects

espindola:
With the code import changes this is now trivial. If everyone is OK
with the design I will write some documentation.

### DIFF
--- a/cli/src/project.rs
+++ b/cli/src/project.rs
@@ -13,6 +13,7 @@ use utils::without_extension;
 const MANIFEST_FILE: &str = "Chisel.toml";
 const TYPES_DIR: &str = "./models";
 const ENDPOINTS_DIR: &str = "./endpoints";
+const LIB_DIR: &str = "./lib";
 const POLICIES_DIR: &str = "./policies";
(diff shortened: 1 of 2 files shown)


MarinPostma:
I'm not sure I understand what issue this is addressing? Are we just adding a `lib` folder to the project scaffold? What is this folder intended for?

espindola:
> I'm not sure I understand what issue this is addressing? Are we just adding a `lib` folder to the project scaffold? What is this folder intended for?

That, and adding a test showing that this is now possible. The possibility of doing this is a result of 182125bd39439ea695cfa4d286af6106bf0cd151. The use of "lib" is mostly a convention. If others agree it is a good one I will update the docs to say that it is now possible to refactor code between multiple endpoints.

MarinPostma (approved):

penberg:
@espindola I think `lib` is a fine convention for sharing code between endpoints. Btw, it's also likely that people will have monorepos with `frontend` (think Next & Gatsby) and `backend` (ChiselStrike) directories. We also need to figure out a good convention for sharing code between the two parts.

PR #1458, opened 2022-06-21
Add a test documenting a bug

espindola:
This shows that we still have some bugs to fix until the types in
chisel.ts are sound.


espindola:
> @espindola can you maybe add this check to some of the existing files? We can perhaps make a lit test file dedicated to select?

done

PR #1461, opened 2022-06-21
chisel.ts: Introduce `jsonToOutput` to clean up the code and fix typing issues.

BearLemma:
This PR finally brings home the typing fixes. It also simplifies the code quite a bit. 

There are still two smaller problems remaining:

1. As a follow-up we can come up with some way to remove the duplication of implementation of most `jsonToOutput` implementations, but I think it can be a follow-up.  
2. The function `jsonToOutput: 
```
jsonToOutput(rawJson: unknown): T {
    return this.inner!.jsonToOutput(rawJson);
}
```
has to use the non-undefined `this.inner!` assertion which might be possible to avoid as the only case where this happens is for `BaseEntity`. It would be nice to have this expressed type-wise, but I don't think it's that big of a problem.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -354,6 +354,11 @@ class SortBy<T> extends Operator<T, T> {
 /** ChiselCursor is a lazy iterator that will be used by ChiselStrike to construct an optimized query. */
 export class ChiselCursor<T> {
     constructor(private inner: Operator<unknown, T>) {}
+
+    // FIXME: The typing of Select operator is wrong because Pick allows to select
+    // not only properties, but methods as well. Which means `Person.cursor().select("save");`
+    // will pass the compiler, but it's obviously not what we want to allow.
+
     /** Force ChiselStrike to fetch just the `...columns` that are part of the colums list. */
     select<C extends (keyof T)[]>(
         ...columns: C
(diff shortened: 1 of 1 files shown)


api/src/chisel.ts
```diff
@@ -216,6 +205,10 @@ class ColumnsSelect<T, C extends (keyof T)[]>
             },
         };
     }
+
+    jsonToOutput(rawJson: unknown): T {
+        return rawJson as T;
```
espindola, line 210: Please add a comment as to why this type assertion is correct. It should basically be the comment lost in runChiselQuery.

api/src/chisel.ts
```diff
@@ -344,45 +342,38 @@ class SortBy<T> extends Operator<T, T> {
...
     }
+
+    jsonToOutput(rawJson: unknown): T {
+        return this.inner!.jsonToOutput(rawJson);
+    }
 }
 
 /** ChiselCursor is a lazy iterator that will be used by ChiselStrike to construct an optimized query. */
 export class ChiselCursor<T> {
-    // FIXME: The types of the various Operators we create is
```
espindola, line 351: This is not fixed, please keep the comment.

api/src/chisel.ts
```diff
@@ -36,6 +35,8 @@ abstract class Operator<Input, Output> {
         iter: AsyncIterable<Input>,
     ): AsyncIterable<Output>;
 
+    public abstract jsonToOutput(rawJson: unknown): Output;
```
espindola, line 38: Nit. Not sure "json" is the right word. C++ has POD. How about record since TS has a Record type?

espindola (approved):

PR #1465, opened 2022-06-22
Remove lib dom

espindola:
This is required for being able to import
https://deno.land/std@0.144.0/node/module.ts which requires
lib.deno_web.d.ts which conflicts with lib.dom.d.ts.
    
We also need to add deno.worker since it defines apis (also defined in
lib.dom.d.ts) used by worker.ts.


espindola:
In draft since it goes on top of https://github.com/chiselstrike/chiselstrike/pull/1463 and https://github.com/chiselstrike/chiselstrike/pull/1462

PR #1473, opened 2022-06-23
migrate lazy_static to one_cell

MarinPostma:
This pr migrates `lazy_static` to `once_cell`. The rational behind that is that once_cell is more idiomatic and easier to use than lazy_static, and doesn't require the use of macros. On top of that there is an effort to introduce in into std: https://doc.rust-lang.org/std/lazy/struct.Lazy.html.

We already transitively depended on this crate, so this does not impact compile time.


espindola:
> This pr migrates `lazy_static` to `once_cell`. The rationale behind that is that once_cell is more idiomatic and easier to use than lazy_static, and doesn't require the use of macros. On top of that, there is an effort to introduce it into std: https://doc.rust-lang.org/std/lazy/struct.Lazy.html.

This would be a nice rational to have in the commit messages. No need to change it now, but for the future, prefer to have info in the commit messages instead of the PR description.

espindola (approved):

PR #1476, opened 2022-06-23
Install cargo tools in isolated path in tests

MarinPostma:
This PR makes the test install the cargo tool in a custom directory, so we don't mess with the developper rust environment.

### DIFF
--- a/cli/tests/linters.rs
+++ b/cli/tests/linters.rs
@@ -10,7 +10,7 @@ mod tests {
     use toml::Value;
 
     fn cargo<'a, T: IntoIterator<Item = &'a str>>(args: T) -> Command {
-        run("cargo", args)
+        run("cargo", args, None)
     }
 
     fn nightly<'a, T: IntoIterator<Item = &'a str>>(args: T) -> Command {
@@ -33,8 +33,8 @@ mod tests {
 
     #[test]
     fn eslint() {
-        run("npm", ["install"]);
-        run("npx", ["eslint", ".", "--ext", ".ts"]);
+        run("npm", ["install"], None);
+        run("npx", ["eslint", ".", "--ext", ".ts"], None);
     }
 
     fn get_deno_version() -> String {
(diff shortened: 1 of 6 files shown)


cli/tests/linters.rs
```diff
@@ -5,12 +5,22 @@ mod common;
...
+    use std::ffi::OsString;
+    use std::fs::{create_dir_all, File};
     use std::io::Read;
+    use std::path::PathBuf;
+    use std::process::Command as StdCommand;
     use toml::Value;
 
+    static CHISEL_BIN_CACHE: &str = ".chisel_dev";
+    static CHISEL_CARGO_PATH: Lazy<OsString> = Lazy::new(|| {
+        let new_path = format!("{}/bin:{}", CHISEL_BIN_CACHE, env!("PATH"));
```
espindola, line 18: Don't we need to put an absolute path in PATH?. We do have a run_in function that changes the current directory.

cli/tests/linters.rs
```diff
@@ -7,14 +7,15 @@ mod tests {
...
     use std::ffi::OsString;
-    use std::fs::File;
+    use std::fs::{create_dir_all, File};
     use std::io::Read;
     use std::path::PathBuf;
+    use std::process::Command as StdCommand;
     use toml::Value;
 
-    static CHISEL_BIN_CACHE: Lazy<PathBuf> = Lazy::new(|| PathBuf::from("./.chisel_dev/bin/"));
+    static CHISEL_BIN_CACHE: &str = ".chisel_dev";
```
espindola, line 16: CHISEL_BIN_CACHE is created in one commit as a `Lazy<PathBuf>` and changed in the next to `&str`. Please create it as a `&str` in the first commit.

cli/tests/linters.rs
```diff
@@ -28,16 +29,43 @@ mod tests {
         ret
     }
 
+    fn is_installed_cargo_pkg(pkg: &str, version: &str) -> bool {
```
espindola, line 32: Why do we now need to check if something is installed? Cargo should take care of that, no?

espindola (commented):
The commit message of a patch says "install and run cargo too". I think you mean tool.

espindola (approved):

Issue #1448, opened 2022-06-20
Support for cyclic dependency in models


MarinPostma:
#1443 fixed a bug in the handling of dependent models. In the doing, we decided to error when cycles were detected in model dependency because we are unable to handle them at the moment.

This issue discusses what would need to be changed to handle cyclic model dependency in the general case.

## Handling cyclic dependency at the type level

The first issue we'll encounter when dealing with cyclic dependency in models in building its `Type` instance (https://github.com/chiselstrike/chiselstrike/blob/main/server/src/types.rs#L487-L494) since `Type` is recursive on the `ObjectType`'s field. We'd need to refactor this code to have some kind of indirection on the  `Type` and avoid direct type recursion.

## Storing entities with cyclic dependencies

This should not be an issue in the relational model, we'd have each table have external references to the entity in the foreign table, no biggie.

## Retrieving entities with a cyclic dependency

To do that we'd need laziness or some other way (some kind of graph traversal?) to prevent infinite recursion when retrieving the object.

espindola:
For retrieval,  we could create a cyclic object, no? If we have models A -> B -> C, one option is to always create a JS object for A, B and C, so they can refer to each other.

MarinPostma:
This is the issue with eagerly loading, @BearLemma is working on collections, I will read his proposal, maybe there is something that could help us here.

MarinPostma:
@espindola yes yes, but you would potentially be eagerly loading a long chain of those objects (e.g a chained list)

espindola:
> @espindola yes yes, but you would potentially be eagerly loading a long chain of those objects (e.g a chained list)

Fair point. I *think* that is OK, since we do offer .select which can be used to fetch only a few fields, and should, module bugs, avoid reading fields that were not requested.

MarinPostma:
If that ever happens, the workaround would be to wrap it in the lazy type @BearLemma is working on for one-to-many relationships

7flash:
That's biggest pain when working with Java JPA models..

PR #1478, opened 2022-06-23
Non-recursive Type

MarinPostma:
The `Field`s of an `Entity` have a `type_: Type` field, which makes the
`Type` being reccursive. This makes it impossible to instanciate types
that that cyclic dependency between them (including self-reference).

This PR makes another step towards supporting self-referential Entity,
and cyclic dependency in Entitities

step towards #1448

### DIFF
--- a/server/src/types.rs
+++ b/server/src/types.rs
@@ -65,15 +65,16 @@ impl VersionTypes {
 }
 
 fn string_field(name: &str) -> Field {
+    let api_version = "__chiselstrike".to_string();
     Field {
         id: None,
         name: name.into(),
-        type_: Type::String,
+        type_id: TypeId::String,
         labels: vec![],
         default: None,
         effective_default: None,
         is_optional: false,
-        api_version: "__chiselstrike".into(),
+        api_version,
         is_unique: false,
     }
 }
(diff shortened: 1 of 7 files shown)


server/src/types.rs
```diff
@@ -671,6 +697,58 @@ impl<'a> ObjectDescriptor for NewObject<'a> {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct TypeId {
```
espindola, line 703: Small preference for making this an enum. We don't need a name or api_version for a float, right?

espindola (commented):
Sounds reasonable. I think I would have a preference to making typeid an enum or really just a uint32 that can be used to index into a type array, but this is probably an improvement as is.

I will defer to @BearLemma since he has been doing most of the work on this area.

server/src/types.rs
MarinPostma: I guess we could work with an enum, I'll try that.

We can discuss the typed being a integer, but I'm not sure how that would work with an intermediary mapping, since will need to reference a type before we can construct it

espindola: Sure, in things that use an index to reference nodes in a cyclic graph, adding a node looks like

* See if it already exist
* Create a placeholder to have a number
* Recurse
* Create the node

MarinPostma: 👍 I think this needs a bit more work on the type system itself. I'll look in that direction in further work if that's ok for you?

server/src/types.rs
```diff
@@ -698,54 +687,46 @@ impl<'a> ObjectDescriptor for NewObject<'a> {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub(crate) struct TypeId {
```
espindola: It is odd to first create TypeId as a struct and then change it to an enum in the following patch. Could you resplit the patches or maybe just squash the first two?

PR #1486, opened 2022-06-28
Added hide transform policy to exclude sensitive or private fields 

notTJ:
Added hide transform policy to exclude sensitive or private fields from the returning json object

### DIFF
--- a/server/src/datastore/query.rs
+++ b/server/src/datastore/query.rs
@@ -52,6 +52,13 @@ impl RequestContext<'_> {
     }
 }
 
+/// Whether a field should be included in or omitted from query result.
+#[derive(Debug, Clone)]
+pub(crate) enum KeepOrOmitField {
+    Keep,
+    Omit,
+}
+
 #[derive(Debug, Clone)]
 pub(crate) enum QueryField {
     Scalar {
@@ -66,7 +73,7 @@ pub(crate) enum QueryField {
         /// Policy transformation to be applied on the resulting JSON value.
         transform: Option<fn(Value) -> Value>,
         /// Do not include field in return json
-        hide: bool,
+        keep_or_omit: KeepOrOmitField,
     },
     Entity {
         /// Name of the original Type field
@@ -75,7 +82,7 @@ pub(crate) enum QueryField {
         /// Policy transformation to be applied on the resulting JSON value.
         transform: Option<fn(Value) -> Value>,
         /// Do not include field in return json
-        hide: bool,
+        keep_or_omit: KeepOrOmitField,
     },
 }
 
@@ -260,7 +267,7 @@ impl QueryPlan {
                 Type::Object(_) => Type::String, // This is actually a foreign key.
                 ty => ty,
             };
-            let field = builder.make_scalar_field(&field, ty.backing_table(), None, false);
+            let field = builder.make_scalar_field(&field, ty.backing_table(), None, &KeepOrOmitField::Keep);
             builder.entity.fields.push(field)
         }
         builder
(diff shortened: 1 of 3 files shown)


notTJ:

notTJ:
recheck

server/src/policies.rs
```diff
@@ -149,6 +156,15 @@ impl VersionPolicy {
                             },
                         );
                     }
+                    Some("hide") => {
```
dekimir, line 159: To avoid confusion with another [proposed](https://chiselstrike.notion.site/New-Policy-Language-for-ChiselStrike-f1720f065a8d4220a113590fd868f7cc) "hide" policy, may I suggest renaming this?  Perhaps "omit" or "omit_field"?

server/src/datastore/query.rs
```diff
@@ -256,7 +260,7 @@ impl QueryPlan {
                 Type::Object(_) => Type::String, // This is actually a foreign key.
                 ty => ty,
             };
-            let field = builder.make_scalar_field(&field, ty.backing_table(), None);
+            let field = builder.make_scalar_field(&field, ty.backing_table(), None, false);
```
dekimir, line 263: `false` isn't very readable here.  Instead of a Boolean, I'd use an enum for this, like in these examples:
https://github.com/chiselstrike/chiselstrike/blob/8c7903548e0198d0ded4b03243e3b894b6d4503e/server/src/types.rs#L670
https://github.com/chiselstrike/chiselstrike/blob/8c7903548e0198d0ded4b03243e3b894b6d4503e/server/src/server.rs#L332-L333

dekimir (changes requested):
Looks good, I just have two suggestions.

server/src/datastore/query.rs
notTJ: I added a new enum "KeepOrOmitField", changed query field property name is "keep_or_omit". Are these names ok?

server/src/policies.rs
```diff
@@ -39,7 +39,7 @@ pub(crate) struct FieldPolicies {
     /// ID of the currently logged-in user.
     pub(crate) current_userid: Option<String>,
     /// Names of fields which will be excluded from query's resulting json object.
-    pub(crate) hide: HashSet<String>,
+    pub(crate) omit: HashSet<String>,
```
notTJ, line 42: Would there be any reason we may specify an additional "Keep" transform, overriding a field which would otherwise be omitted? This could be a hashmap of <String, KeepOrOmitField>. I'm kind of wondering if there's a better name than just omit... "omitted_fields"?

dekimir: > Would there be any reason we may specify an additional "Keep" transform in the new policy language? It would include a field which would otherwise be omitted. This could be a hashmap of <String, KeepOrOmitField>.

That would make conflicting policies possible, right?  We'd need to specify how those conflicts are resolved.  Sounds complex.  I'd avoid that until it's deemed necessary by enough potential users.

> I'm also wondering if there's a better name than just omit... "omitted_fields"?

Yes, that's better.  But it can be a follow-up.

dekimir (dismissed):
This is good enough for now.  I'll create a follow-up with tests and documentation.  If that reveals any problems, we'll fix them in stride.

glommer:
hey there! FYI: CI are not passing, because we enforce coding style and clean linters.

Click "details" to see what exactly is wrong, but running `cargo fmt --all` and `cargo clippy` should fix it!

dekimir:
@notTJ looks like you still need to run `cargo clippy`:
https://github.com/chiselstrike/chiselstrike/runs/7114021755?check_suite_focus=true#step:14:424

notTJ:
> 

my bad I thought it worked like fmt --all, it should pass now

glommer (approved):

glommer:
Thanks!!!

PR #1488, opened 2022-06-28
Remove QueryEngine::mutate method

### LINKED ISSUE #1487: Remove QueryEngine::mutate method
Following discussion with @BearLemma and @espindola, as well as changes from #1480, it would be desirable to remove the `QueryEngine::mutate`, and have all mutations on the query engine explicitly take a transaction. This will hopefully prevent bugs like #1482

MarinPostma:
close #1487

### DIFF
--- a/server/src/datastore/engine.rs
+++ b/server/src/datastore/engine.rs
@@ -480,6 +480,9 @@ impl QueryEngine {
     }
 
     /// Execute the given `mutation`.
+    ///
+    /// Only for testing purposes. For any other purpose, use `mutate_with_transaction`.
+    #[cfg(test)]
     pub(crate) async fn mutate(&self, mutation: Mutation) -> Result<()> {
         let mut transaction = self.start_transaction().await?;
         let raw_sql = mutation.build_sql(self.target_db())?;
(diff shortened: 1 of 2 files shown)


server/src/deno.rs
```diff
@@ -592,7 +592,19 @@ async fn op_chisel_crud_delete(
...
+
+    let transaction = query_engine.clone().start_transaction_static().await?;
+
+    let mut guard = transaction.lock().await;
+    query_engine
+        .mutate_with_transaction(mutation, &mut guard)
+        .await?;
+
+    drop(guard);
+
```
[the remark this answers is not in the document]
MarinPostma: I like 'em

server/src/datastore/engine.rs
```diff
@@ -480,6 +480,9 @@ impl QueryEngine {
     }
 
     /// Execute the given `mutation`.
+    ///
+    /// Only for testing purposes. For any other purpose, use `mutate_with_transaction`.
+    #[cfg(test)]
```
espindola, line 485: I would either move this next to the tests or change the tests to use mutate_with_transaction.

MarinPostma: The tests are spread across different files, which is why I left them here. We can also put it in this file but in a test module?

PR #1489, opened 2022-06-28
Introduce `Entity` type covering Custom entity type and Auth types.

BearLemma:
This PR aims to make our type system more expressive and paves a way for extended typing support (`List`, `Array`, etc.)

There are multiple patches so reviewers are advised to go patch by patch. 
First thing this PR does is it introduces `Type::Entity(Entity)` and standalone `enum Entity {Auth, Custom}` types and makes amendments where necessary for the code to compile. 
Then it renames `add_builtin_object_type` -> `add_auth_type` and `add_type` -> `add_custom_type`.
Finally it brings home the idea and removes the `is_auth` field from `ObjectType` in favor of `Entity::is_auth` convenience method.

### DIFF
--- a/server/src/deno.rs
+++ b/server/src/deno.rs
@@ -506,7 +506,7 @@ async fn op_chisel_store(
     let (query_engine, ty) = {
         let state = state.borrow();
         let ty = match current_type_system(&state).lookup_type(type_name, &c.api_version) {
-            Ok(Type::Object(ty)) => ty,
+            Ok(Type::Entity(ty)) => ty,
             _ => anyhow::bail!("Cannot save into type {}.", type_name),
         };
         if ty.is_auth() && !is_auth_path(&c.api_version, &c.path) {
--- a/server/src/types.rs
+++ b/server/src/types.rs
@@ -3,7 +3,6 @@
 use crate::auth::{AUTH_ACCOUNT_NAME, AUTH_SESSION_NAME, AUTH_TOKEN_NAME, AUTH_USER_NAME};
 use crate::datastore::query::{truncate_identifier, QueryPlan};
 use crate::datastore::QueryEngine;
-use crate::types::AuthOrNot::IsAuth;
 use anyhow::Context;
 use deno_core::futures;
 use derive_new::new;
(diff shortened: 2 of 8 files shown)


server/src/deno.rs
```diff
@@ -506,8 +506,12 @@ async fn op_chisel_store(
...
-            Ok(Type::Object(ty)) => ty,
+        let ty = current_type_system(&state)
+            .lookup_type_primitive(type_name, &c.api_version)
+            .with_context(|| format!("trying to save into an unknown type `{type_name}`"))?;
+
+        let ty = match ty {
+            Primitive::Object(ty) => ty,
             _ => anyhow::bail!("Cannot save into type {}.", type_name),
         };
         if ty.is_auth() && !is_auth_path(&c.api_version, &c.path) {
```
dekimir, line 517: It feels like `Type::is_auth` should be affected by this change, yet it isn't.

dekimir (changes requested):
I don't understand how adding Primitive is related to singling out Auth types.  I also don't know what the alternative to primitive types is.  (For example: the alternative to scalar types is entities, and the alternative to builtin types is user-defined ones.)  Perhaps it's just a matter of explaining it in the commit message?

dekimir (approved):
Approving to allow progress on List fields.  But I'm not convinced that separating auth types into their own enum elements is well-advised; they're more similar than different, and the separation doesn't extend to their ObjectType representations (eg: `get_auth_user_type()` looks up `AuthUser` by its name, even after we made it a unique known enum).  I'd like to see that commit reverted one day.

I also dislike the ambiguous and vague name "primitive".  These are really "storage" or "storable" types.

server/src/types.rs
```diff
@@ -371,6 +386,13 @@ impl TypeSystem {
             .ok_or_else(|| TypeSystemError::NotABuiltinType(type_name.to_string()))
     }
 
+    pub(crate) fn lookup_builtin_primitive(
+        &self,
+        type_name: &str,
+    ) -> Result<Primitive, TypeSystemError> {
+        self.lookup_builtin_type(type_name).map(|t| t.primitive())
+    }
+
```
MarinPostma, line 395: nitpick: I don't think that these method (along with `lookup_type_primitive`) are very helpful. Calling the type counterpart + `.primitive()` is a lot more explicit to me, since this `type`-`primitive` thing is already quite confusing

dekimir (approved):
Very nice job of using digestible, logical commits! Let's always do that. :D

Issue #1494, opened 2022-06-28
server arguments are not passed through `chisel`


glommer:
If we invoke the server through `chisel start` or `chisel dev`, the server arguments are not passed to the server at all.
That's how we document usage, meaning we should pass options like port selection to the server if the server is invoked through the `chisel` binary.

Found by user @7Flash

See #1479

7flash:
Working on PR.. Is that okay just to accept "internal_routes_listen_addr" in CLI and pass as argument to start_server()..

glommer:
I think the best would be to just passthrough all options. There are many other options in the server that are relevant as well. This is also more future proof, because when more options are added they are passed as well.

`std::process::Command` has an `args` parameter, so I think the best is to just read all existing arguments passed and pass them along.

Issue #1481, opened 2022-06-27
Passing argument from cli to server on launch


MarinPostma:
Right now, I want to develop with some custom server parameters (I'm mostly thinking of changing the address and port to which it serves), and I need to spin an independent chiseld instance. This is not very user-friendly. It would be handy to pass arguments to the cli, that are passed to the server. On the top of my head, I'm thinking that any arguments passed after `--` could be passed when launching the server.

Also considering that passing any argument to the server could put it in a state that makes it impossible to operate with the cli out of the box...

could be related to https://github.com/chiselstrike/chiselstrike/issues/1479

penberg:
@MarinPostma This is fixed now, isn't it?

MarinPostma:
yes, I think it is

MarinPostma:
closed by #1521

PR #1495, opened 2022-06-29
passing port argument from CLI to server  #1494

7flash:


7flash:
The following changes has been made..

First.. port argument is accepted in start command.. wildcard is used to ignore the actual value..

image

Second.. all arguments are passed from CLI to the server command.. 

image

7flash:
Consider #1 to pass other args.. Consider #2 to do the same for Dev command..

7flash:
Testing:

```
cd my-backend
RUST_LOG=debug ../target/debug/chisel start --internal-routes-listen-addr=127.0.0.1:9095
```

image

7flash:
https://github.com/chiselstrike/chiselstrike/issues/1481 https://github.com/chiselstrike/chiselstrike/issues/1494

7flash:
merge can be blocked by https://github.com/chiselstrike/chiselstrike/issues/1496

dekimir:
Thank you for this, @7flash.  May I ask that you provide a description in the commit message rather than the PR comments.  You should also put "Fixes #1494" in there, so GitHub connects the two.

An easy way to edit your commit message is `git commit --amend` from the command line in your `cli-args` branch.

cli/src/main.rs
```diff
@@ -78,7 +79,10 @@ enum Command {
         auto_index: bool,
     },
     /// Start the ChiselStrike server.
-    Start,
+    Start {
+        #[structopt(short, long, default_value="127.0.0.1:9090")]
+        internal_routes_listen_addr: SocketAddr
```
dekimir, line 84: Why is this added?  It doesn't seem to be used anywhere.

dekimir (commented):
I wonder what would be the best way to test this?  Can be another PR later.

cli/src/main.rs
7flash: otherwise getting an error "error[E0027]: pattern does not mention field `internal_routes_listen_addr`"..

7flash: note that currently it would give similar error when trying to pass "chisel --rpc_listen_addr".. all supported arguments should be defined here..

dekimir: Ah, OK.  Then I suggest using [external_subcommand](https://docs.rs/structopt/latest/structopt/index.html#external-subcommands) here, to eliminate such errors.

7flash:
@dekimir regarding your suggestion.. external_subcommand throws "enums-only" error..

image

image

7flash:
Nevermind.. it works when applied properly..

cli/src/server.rs
```diff
@@ -26,15 +26,17 @@ pub(crate) fn start_server() -> anyhow::Result<std::process::Child> {
...
     cmd.pop();
     cmd.push("chiseld");
-    let server = match std::process::Command::new(cmd.clone()).spawn() {
-        Ok(server) => server,
-        Err(e) => {
-            match e.kind() {
-                ErrorKind::NotFound => anyhow::bail!("Unable to start the server because `chiseld` program is missing. Please make sure `chiseld` is installed in {}", cmd.display()),
-                _ => anyhow::bail!("Unable to start `chiseld` program: {}", e),
+    let server = match std::process::Command::new(cmd.clone())
+        .args(std::env::args().skip(2))
```
dekimir, line 30: Nitpick: this is fragile.  It will break if `start` doesn't immediately follow `chisel`, which is allowed (though currently unlikely to happen -- but we don't know what the future brings).

dekimir (approved):
Looks good.  The one nitpick I have can be addressed in a follow-up.

It also needs a test, which we can work on later.

PR #1503, opened 2022-06-30
Introduce more expressive type support in protobuf between client and server

BearLemma:
Next PR in the `List` odyssey. 
This PR introduces new protobuf type 
```
message TypeMsg {
  oneof type_enum {
    bool string = 1;
    bool number = 2;
    bool bool = 3;
    string entity = 4;
  };
}
``` 
which represents various type flavors which map nicely to server's `Type` enum. I tried splitting the PR into multiple patches, but unfortunately, the main bulk of changes needs to go all at once to make the for a nice standalone commit.

### DIFF
--- a/cli/src/ts.rs
+++ b/cli/src/ts.rs
@@ -1,7 +1,8 @@
-use crate::chisel::{AddTypeRequest, FieldDefinition};
+use crate::chisel::{type_msg::TypeEnum, AddTypeRequest, FieldDefinition, TypeMsg};
 use anyhow::{anyhow, bail, ensure, Context, Result};
 use chisel_server::auth::is_auth_entity_name;
 use std::collections::BTreeSet;
+use std::fmt;
 use std::path::Path;
 use swc_common::sync::Lrc;
 use swc_common::{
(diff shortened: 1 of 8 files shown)


cli/src/ts.rs
```diff
@@ -142,15 +143,15 @@ fn get_type_decorators(handler: &Handler, x: &[Decorator]) -> Result<(Vec<String
...
 fn validate_type_vec(type_vec: &[AddTypeRequest], valid_types: &BTreeSet<String>) -> Result<()> {
-    let mut builtin_types: BTreeSet<&str> = BTreeSet::new();
-    builtin_types.insert("string");
-    builtin_types.insert("number");
-    builtin_types.insert("boolean");
-    builtin_types.insert("AuthUser");
+    let mut scalar_types: BTreeSet<&str> = BTreeSet::new();
+    scalar_types.insert("string");
+    scalar_types.insert("number");
+    scalar_types.insert("boolean");
```
MarinPostma:
```suggestion
static SCALAR_TYPES: Lazy<BTreeSet<&str>>  = Lazy::new(|| ["string", "number", "boolean"].into_iter().collect());
```

server/src/rpc.rs
```diff
@@ -559,6 +567,51 @@ fn sort_custom_types(
...
+            Type::Id => TypeEnum::String(true),
+            Type::String => TypeEnum::String(true),
+            Type::Boolean => TypeEnum::Bool(true),
+            Type::Entity(entity) => TypeEnum::Entity(entity.name().to_owned()),
+        };
+        TypeMsg {
+            type_enum: Some(ty),
+        }
+    }
+}
```
MarinPostma, line 613: That calls for an `impl From<Type> for TypeMsg`

server/src/rpc.rs
```diff
@@ -533,9 +536,14 @@ fn sort_custom_types(
...
-                graph.add_node(field.field_type.as_str());
-                graph.add_edge(field.field_type.as_str(), ty.name.as_str(), ());
+            let field_type = field.field_type()?;
+            if !field_type.is_builtin(ts) {
+                let entity_name = match field_type {
+                    TypeEnum::Entity(name) => name,
+                    _ => anyhow::bail!("if type is not a builtin, it must be an entity"),
+                };
+                graph.add_node(entity_name);
+                graph.add_edge(entity_name, ty.name.as_str(), ());
```
MarinPostma, line 546: Actually, the error branch should be unreachable, I think it's cleaner this way:
```suggestion
match field_type {
    TypeEnum::Entity(name) if field_type.is_builtin(ts) => {
                graph.add_node(entity_name);
                graph.add_edge(entity_name, ty.name.as_str(), ());
    }
    _ => ()
```

MarinPostma (approved):

PR #1505, opened 2022-06-30
Minor code changes to make code compilable

NickHodges:
`.ts` extension is not allowed on import references

### DIFF
--- a/docusaurus/docs/Intro/endpoints.md
+++ b/docusaurus/docs/Intro/endpoints.md
@@ -14,7 +14,7 @@ with either `{"data": VALUE}` or `{"error": "message"}` depending on the result.
 
 ```typescript title="my-backend/endpoints/comments-readonly.ts"
 import { crud, standardCRUDMethods, responseFromJson } from "@chiselstrike/api";
-import { BlogComment } from "../models/BlogComment.ts";
+import { BlogComment } from "../models/BlogComment";
 export default crud(
     BlogComment,
     ":id", /* :id can be explicitly provided */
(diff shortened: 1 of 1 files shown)


ricardodalarme (approved):

docusaurus/docs/Intro/endpoints.md
```diff
@@ -14,7 +14,7 @@ with either `{"data": VALUE}` or `{"error": "message"}` depending on the result.
 
 ```typescript title="my-backend/endpoints/comments-readonly.ts"
 import { crud, standardCRUDMethods, responseFromJson } from "@chiselstrike/api";
-import { BlogComment } from "../models/BlogComment.ts";
+import { BlogComment } from "../models/BlogComment";
```
dekimir, line 17: Interestingly enough, even the old version (with `.ts`) works fine for me.  @NickHodges what error message do you get?  And what platform are you on?  (I'm on Linux, using npm v8.3.0.)

espindola: I don't think this is more or less compatible, it is just a different mode. In deno mode you must use the .ts, in node mode you cannot. Deno requires extensions because browsers do (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#importing_features_into_your_script).

If your are seeing issues with vscode, note that it don't support importing URLs or using .ts extensions. So deno offers a vscode extension for better development experience: https://deno.land/manual/vscode_deno

dekimir: The documentation doesn't currently mention the Deno mode.  It uses Node mode exclusively.

dekimir: Hm, the above VSCode error remains even after I turn on the Deno extension. :thinking:

PR #1509, opened 2022-07-01
chiseld config file

MarinPostma:
On chiseld startup, attempts to read the config file from the platform specific
config directory, or from the file pointed to by the `-c` cli argument.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -12,7 +12,14 @@ use server::DoRepeat;
 use std::env;
 use std::ffi::CString;
 use std::io::Write;
+use std::path::PathBuf;
 use structopt::StructOpt;
+
+fn find_default_config_path() -> Option<PathBuf> {
+    let config_dir = dirs::config_dir()?.join("chiselstrike");
+    let config_path = config_dir.join("config.toml");
+    config_path.exists().then(|| config_path)
+}
 
 #[tokio::main]
 async fn main() -> Result<()> {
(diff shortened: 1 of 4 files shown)


server/src/server.rs
```diff
@@ -66,6 +69,18 @@ pub struct Opt {
     /// If on, serve a web UI on an internal route.
     #[structopt(long)]
     webui: bool,
+    /// Run chiseld from toml configuration file
```
dekimir, line 72: "Run chiseld from" reads unclear to me.  Consider something like "Read chiseld options from a config file instead of the command line."

dekimir (commented):
I like this, and it's elegantly implemented.

server/src/server.rs
dekimir: Sounds good.

It took some reading for me to understand what the order of priority is.  Do I understand correctly that the default config file can specify a non-default config file that completely overrides it?  (Except that the non-default config file's `config` variable is ignored.)

MarinPostma: Yes that's correct. I initially thought about disallowing using this parameter from within a configuration file, that's feasible too.

dekimir: > Yes that's correct. I initially thought about disallowing using this parameter from within a configuration file, that's feasible too.

I like that.  It seems cleaner and less complicated to document.

server/src/main.rs
```diff
@@ -32,7 +39,20 @@ async fn main() -> Result<()> {
...
 
-    if let DoRepeat::Yes = server::run_all(server::Opt::from_args()).await? {
+    let opt = {
+        let default_path = find_default_config_path();
+        let opt = match default_path {
+            Some(ref path) => server::Opt::from_file(path).await?,
+            None => server::Opt::from_args(),
+        };
+
+        match opt.config {
```
dekimir, line 49: I thought we'll ignore this setting if `opt` came `from_file`?

dekimir (approved):
Approving for forward progress, but:
1. Needs tests (in a follow-up)
2. Let's implement the suggestion from https://github.com/chiselstrike/chiselstrike/pull/1509#discussion_r913277546

server/src/main.rs
MarinPostma: @dekimir, we disallow setting the `from_file` in the config file altogether: https://github.com/chiselstrike/chiselstrike/pull/1509/files#diff-e4e539f1719bacc61407b3ccd49ec44b594e30e2509f6a76bba6e65c4387e40bR74

It's probably a bit trickier to check if this option was actually coming from the files, because the configuration is getting merged by the library. (not impossible though, but probably hacky)

dekimir: Ah, OK.

dekimir (approved):

PR #1521, opened 2022-07-06
pass cli args to chiseld

MarinPostma:
Pass any cli arguments after a `--` to chiseld on  `start` and `dev`
command, e.g:

`chiseld dev -- --help` : prints chiseld help

Followup: 
- I need another PR to make chiselstrike correctly exit when chiseld exits.
- Add test
- Add documentation

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -79,8 +79,7 @@ enum Command {
         auto_index: bool,
     },
     /// Start the ChiselStrike server.
-    #[structopt(external_subcommand)]
-    Start(Vec<String>),
+    Start,
     /// Show ChiselStrike server status.
     Status,
     /// Restart the running ChiselStrike server.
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -256,7 +263,7 @@ async fn main() -> Result<()> {
             create_project(path, opts)?;
         }
         Command::Start { .. } => {
```
espindola, line 265: This is now just Command::Start, right?

cli/src/main.rs
```diff
@@ -151,7 +150,15 @@ pub(crate) async fn restart(server_url: String) -> Result<()> {
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    let opt = Opt::from_args();
+    let chisel_args = std::env::args().take_while(|arg| arg != "--");
```
espindola, line 153: Somewhat shocked that there is no builtin way to do this with structopt, but I could not find one.

espindola (approved):
LGTM with a possible nit for a followup.

espindola (approved):

PR #1522, opened 2022-07-06
Warn about initialization of nested entities and fix shallow object.assign

BearLemma:
Fixes the issue of replacing nested entities with plain records.

### DIFF
--- a/cli/src/ts.rs
+++ b/cli/src/ts.rs
@@ -112,6 +112,8 @@ fn get_field_value(handler: &Handler, x: &Option<Box<Expr>>) -> Result<Option<(S
                     .ok_or_else(|| swc_err(handler, k, "unexpected empty expression"))?;
                 Ok(Some((format!("{}{}", op, value.0), value.1)))
             }
+            // Covers entity fields initialization.
+            Expr::New(_) => Ok(None),
             x => Err(swc_err(handler, x, "expression not supported")),
         },
     }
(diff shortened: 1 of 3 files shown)


cli/tests/lit/store-nested-type.deno
```diff
@@ -109,3 +109,29 @@ $CURL -X POST $CHISELD_HOST/dev/update
...
+    await company!.ceo!.save();
+
+    return new Response('Successfully updated');
+}
+EOF
+
+$CHISEL apply
+
+$CURL -X POST $CHISELD_HOST/dev/update
+# CHECK: Successfully updated
```
espindola, line 133: Would you like to change the test to CHECK the current failure and have a FIXME saying what needs to change? That way the PR can be merged.

dekimir: Or `Person.create(company!.ceo!) // TODO: replace create() with save()`, and keep the current CHECK.

cli/tests/lit/savewrong.deno
```diff
@@ -65,7 +65,7 @@ cat << EOF > "$TEMPDIR/models/post.ts"
 import { ChiselEntity, AuthUser } from '@chiselstrike/api'
 export class Post extends ChiselEntity {
     text: string = "Lorem Ipsum";
-    author: AuthUser;
+    author: AuthUser = new AuthUser();
```
dekimir, line 68: Is this required?  If so, that's not a good solution.

dekimir: I don't think this is an acceptable user experience.

dekimir: Correct me if I'm wrong, but isn't the "dysfunction" just `.save()` not working on subobjects?  That's less bad than forcing every user to mindlessly type `= new Foo()` after every nested field.  On rare occasions that you need to save subobjects, there's an easy workaround with `Foo.create()`, right?

dekimir: > Regarding your workaround with Foo.create(), that doesn't work either and has the exact same problems.

Are we talking about the same thing?  I meant calling `Foo.create()` on the subobject, not the root object.  I can't imagine that not working; the return type of `await Foo.create()` is Foo, with all its methods.  The root object (let's call its type Bar) is not created, it's loaded via, eg, `Bar.findOne()`.

penberg: FWIW, I discussed this privately with @BearLemma, and I think there's a strong argument for _temporarily_ requiring developers to write that explicit property assignment, and then lifting that restriction later with reflection, because the overall developer experience is better. With @dekimir's suggested workaround, the problem is that people will first hit the limitaiton, and then they have to find out about the workaround. With @BearLemma's pull request, you get an error message if you forget to write the explicit assignment.

So my proposal is that instead of _erroring_ we print a warning if there's no assignment. That preserves compatibility and makes the workaround discoverable. Once the pull request is in, we work towards either reflection or doing the assignment automatically with bit of `chiselc` magic.

cli/src/ts.rs
```diff
@@ -186,6 +188,29 @@ fn parse_class_prop(x: &ClassProp, class_name: &str, handler: &Handler) -> Resul
...
+                        \tFor further information, please see https://github.com/chiselstrike/chiselstrike/issues/1541"
+                    );
+            }
+            Some(k) => match &**k {
+                Expr::New(_) => {}
+                x => anyhow::bail!(swc_err(
+                    handler,
+                    x,
+                    &format!(
+                        "field `{field_name}` of entity type `{name}` has unexpected initializer"
```
dekimir, line 206: This error and the above warning are never tested.

dekimir (approved):

cli/tests/lit/savewrong.deno
dekimir: This is acceptable as a temporary solution.

PR #1523, opened 2022-07-06
spawn chiseld with tokio

MarinPostma:
Spawn chiseld command with tokio instead of std.

This allows us to early exit when chiseld exit.

This is necessary for cases where chiseld would exit while chisel is still waiting for it to go up:
```
chisel start -- --help
```

in this case, we were waiting for the server url, before checking if it had exited (https://github.com/chiselstrike/chiselstrike/pull/1523/files#diff-37970e7d817d4316a093bb967d74ff7e26d9c6c01d93f5a79d3d195ba9f6155bL265), basically blocking the process.

followup to #1521

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -626,6 +626,7 @@ version = "0.11.0-dev.0"
 dependencies = [
  "anyhow",
  "endpoint_tsc",
+ "futures",
  "glob",
  "handlebars",
  "itertools 0.10.3",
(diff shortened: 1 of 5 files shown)


espindola:
Can you expand the description a bit? Is this for handling the server crashing while using chisel dev?

MarinPostma:
@espindola done, lmk if that still isn't clear

PR #1534, opened 2022-07-07
api improvements: allow users to return arbitrary types

glommer:
We mandate a Response to be returned, meaning that users have to do
returnFromJson, or in other ways construct responses.

However in many cases, specialized endpoints want to return well
defined things like an array of models, a model, or even strings.

Allow that to happen, by converting to a Response internally.
This is a major UX boost that allow us to craft much shorter endpoints

### DIFF
--- a/cli/tests/lit/return.deno
+++ b/cli/tests/lit/return.deno
@@ -1,0 +1,51 @@
+# SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
+
+# RUN: sh -e @file
+
+cd "$TEMPDIR"
+
+cat << EOF > "$TEMPDIR/endpoints/return.ts"
+type MyReturn = {
+    a: string;
+    b: number;
+}
+
+export default async function() : Promise<MyReturn> {
+    return { "a": "x", "b": 0 };
+}
+EOF
+
+cat << EOF > "$TEMPDIR/endpoints/undefined.ts"
+type MyReturn = {
+    a: string;
+    b: number;
+}
+
+export default async function(req: Request) : Promise<MyReturn | undefined> {
+    return undefined;
+}
+EOF
+
+$CHISEL apply
+
+$CURL $CHISELD_HOST/dev/return
+# CHECK: "a": "x"
+# CHECK: "b": 0
+
+$CURL $CHISELD_HOST/dev/undefined && echo "UniqueString"
+# CHECK: HTTP/1.1 200 OK
+# CHECK UniqueString
+
+cat << EOF > "$TEMPDIR/endpoints/return.ts"
+type MyReturn = {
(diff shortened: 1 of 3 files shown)


docusaurus/docs/Intro/endpoints.md
```diff
@@ -42,10 +42,35 @@ You may also be interested in the [Authentication](InDepth/login.md) chapter.
...
+
+```typescript title="my-backend/endpoints/allcomments.ts"
+import { BlogComment } from "../models/BlogComment"
+
+export default async function() : Promise<Array<BlogComment>> {
+    return BlogComment.findAll()
+}
+```
+
+Optionally, your function can that take as a parameter a `ChiselRequest` or its parent class [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) with a few convenience fields added. In the following example, we can use the `pathComponents()` method of `ChiselRequest` to implement path-based finding:
```
dekimir, line 61: With "or its parent class" this now doesn't make sense.  It reads like the "few convenience fields" are added to the function, not to `ChiselRequest`.

docusaurus/docs/Intro/endpoints.md
```diff
@@ -101,7 +129,7 @@ we were able to parse the request under `POST` with our own custom validation, a
 the `build` API to construct an object that is then persisted with `save`.  We'll explain the use of the
 data model more in [Data Access](Intro/data-access).
 
-Finally, notice how we can return a standard `Response` in some cases, but also can also use the convenience method
+Finally, notice how in the cases where we want to return a `Response`, we can either return a standard `Response` but also the convenience method
```
dekimir, line 132: I've heard of "either/or" but not of "either/but also". :p

cli/tests/lit/return.deno
```diff
@@ -0,0 +1,50 @@
...
+EOF
+
+$CHISEL apply
+
+$CURL $CHISELD_HOST/dev/return
+# CHECK: "a": "x"
+# CHECK: "b": 0
+
+$CURL $CHISELD_HOST/dev/undefined
+# CHECK: HTTP/1.1 200 OK
```
dekimir, line 36: This doesn't ensure that the endpoint returns nothing, as `lit` ignores extra output between CHECKs.  Perhaps you can tack on `&& echo UniqueString` to line 35 and CHECK for UniqueString here?

dekimir (changes requested):

docusaurus/docs/Intro/endpoints.md
glommer: fixed

glommer:
@dekimir rebased

dekimir (approved):
I pushed two small doc fixes, and now this LGTM.

PR #1548, opened 2022-07-11
Update Cargo.lock

espindola:
Sorry about forgetting it in the previous PR.


espindola:
> @espindola how come the pipeline passes when the cargo-lock wasn't in sync with the updated deno submodule?

cargo build updates it.

PR #1552, opened 2022-07-13
Enable mergify queues

penberg:


MarinPostma (commented):
can we enable speculative-checks as well? https://docs.mergify.com/actions/queue/#speculative-checks

penberg:
Yeah, happy to enable it, but in a different pull request to avoid testing two features at the same time.

MarinPostma (approved):

PR #1556, opened 2022-07-13
Improve error messages

honzasp:
While trying to debug imports, I found some confusing error messages. This PR aims to improve them.

### DIFF
--- a/cli/src/cmd/apply/node.rs
+++ b/cli/src/cmd/apply/node.rs
@@ -113,7 +113,8 @@ pub(crate) async fn apply(
     if !res.status.success() {
         let out = String::from_utf8(res.stdout).expect("command output not utf-8");
         let err = String::from_utf8(res.stderr).expect("command output not utf-8");
-        return Err(anyhow!("compiling endpoints")).with_context(|| format!("{}\n{}", out, err));
+        return Err(anyhow!("{}\n{}", out, err))
+            .context("could not bundle endpoints with esbuild (using node-style modules)");
     }
 
     for (endpoint_name, bundler_info) in endpoints.iter().zip(bundler_file_mapping.iter()) {
@@ -141,8 +142,8 @@ pub(crate) async fn apply(
     Ok((endpoints_req, index_candidates_req))
 }
 
-fn npx<A: AsRef<OsStr>, C: AsRef<OsStr>>(
-    command: C,
+fn npx<A: AsRef<OsStr>>(
+    command: &'static str,
     args: &[A],
     stdin: Option<std::process::ChildStdout>,
 ) -> JoinHandle<Result<std::process::Output>> {
@@ -155,6 +156,6 @@ fn npx<A: AsRef<OsStr>, C: AsRef<OsStr>>(
 
     spawn_blocking(move || {
         cmd.output()
-            .with_context(|| "trying to execute `npx esbuild`. Is npx on your PATH?".to_string())
+            .with_context(|| format!("could not execute `npx {}`. Is npx on your PATH?", command))
     })
 }
--- a/cli/tests/lit/bad-url.deno
+++ b/cli/tests/lit/bad-url.deno
@@ -13,6 +13,6 @@ EOF
 cd "$TEMPDIR"
 $CHISEL apply 2>&1 || true
 
-# CHECK: Error: parsing endpoints
+# CHECK: Error: could not compile endpoints
 # CHECK: Caused by:
 # CHECK: HTTP request failed
--- a/cli/tests/lit/http-import.node
+++ b/cli/tests/lit/http-import.node
@@ -1,0 +1,17 @@
+# SPDX-FileCopyrightText: © 2021 ChiselStrike <info@chiselstrike.com>
+
+# RUN: sh -e @file
+
+cat << EOF > "$TEMPDIR/endpoints/error.ts"
+import { foo } from "https://foo.bar";
+
+export default async function chisel(req: Request) {
+    return foo;
+}
+EOF
+
+cd "$TEMPDIR"
+
+$CHISEL apply 2>&1 || true
+# CHECK: could not import endpoint code into the runtime
+# CHECK: chiseld cannot load module https://foo.bar/ at runtime
--- a/server/src/datastore/dbconn.rs
+++ b/server/src/datastore/dbconn.rs
@@ -47,7 +47,7 @@ impl DbConnection {
             .max_connections(nr_conn as _)
             .connect(uri)
             .await
-            .with_context(|| format!("connecting to {}", uri))?;
+            .with_context(|| format!("failed to connect to {}", uri))?;
 
         let conn_uri = uri.to_owned();
 
--- a/server/src/datastore/meta/mod.rs
+++ b/server/src/datastore/meta/mod.rs
@@ -36,7 +36,7 @@ async fn execute<'a, 'b>(
     transaction
         .execute(query)
         .await
-        .with_context(|| format!("Executing query {}", qstr))
+        .with_context(|| format!("Failed to execute query {}", qstr))
 }
 
 async fn fetch_one<'a, 'b>(
@@ -47,7 +47,7 @@ async fn fetch_one<'a, 'b>(
     transaction
         .fetch_one(query)
         .await
-        .with_context(|| format!("Executing query {}", qstr))
+        .with_context(|| format!("Failed to execute query {}", qstr))
 }
 
 async fn fetch_all<'a, E>(
@@ -61,7 +61,7 @@ where
     query
         .fetch_all(executor)
         .await
-        .with_context(|| format!("Executing query {}", qstr))
+        .with_context(|| format!("Failed to execute query {}", qstr))
 }
 
 async fn file_exists(file: &Path) -> anyhow::Result<bool> {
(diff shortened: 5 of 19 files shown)


server/src/datastore/dbconn.rs
```diff
@@ -47,7 +47,7 @@ impl DbConnection {
             .max_connections(nr_conn as _)
             .connect(uri)
             .await
-            .with_context(|| format!("connecting to {}", uri))?;
+            .with_context(|| format!("failed to connect to {}", uri))?;
```
MarinPostma, line 50:
```suggestion
            .with_context(|| format!("failed to connect to {uri}"))?;
```

server/src/datastore/meta/mod.rs
```diff
@@ -36,7 +36,7 @@ async fn execute<'a, 'b>(
     transaction
         .execute(query)
         .await
-        .with_context(|| format!("Executing query {}", qstr))
+        .with_context(|| format!("Failed to execute query {}", qstr))
```
MarinPostma, line 39:
```suggestion
        .with_context(|| format!("Failed to execute query {qstr}"))
```

server/src/datastore/meta/mod.rs
```diff
@@ -47,7 +47,7 @@ async fn fetch_one<'a, 'b>(
     transaction
         .fetch_one(query)
         .await
-        .with_context(|| format!("Executing query {}", qstr))
+        .with_context(|| format!("Failed to execute query {}", qstr))
```
MarinPostma, line 50:
```suggestion
        .with_context(|| format!("Failed to execute query {qstr}"))
```

server/src/datastore/meta/mod.rs
```diff
@@ -61,7 +61,7 @@ where
     query
         .fetch_all(executor)
         .await
-        .with_context(|| format!("Executing query {}", qstr))
+        .with_context(|| format!("Failed to execute query {}", qstr))
```
MarinPostma, line 64:
```suggestion
        .with_context(|| format!("Failed to execute query {qstr}"))
```

tsc_compile/src/lib.rs
```diff
@@ -223,17 +224,17 @@ struct ModuleLoader {
 static ROOT_URL: &str = "chisel://root_domain/root.ts";
 
 fn load_url(extra_libs: &HashMap<Url, String>, specifier: Url) -> impl Future<Output = LoadResult> {
-    let sync_text = match specifier.scheme() {
-        "file" => fs::read_to_string(specifier.to_file_path().unwrap()),
-        "chisel" => Ok(extra_libs.get(&specifier).unwrap().clone()),
-        _ => Ok("".to_string()),
+    let sync_text: Option<Result<String>> = match specifier.scheme() {
+        "file" => Some(fs::read_to_string(specifier.to_file_path().unwrap()).map_err(|err| anyhow!(err))),
```
MarinPostma, line 228: can't we do that?
```suggestion
        "file" => Some(fs::read_to_string(specifier.to_file_path().unwrap()).into()),
```

tsc_compile/src/lib.rs
```diff
@@ -365,7 +366,18 @@ impl Compiler {
         )
         .await;
 
-        graph.valid()?;
+        graph.valid().map_err(|err| {
+            match err {
+                ModuleGraphError::LoadingErr(specifier, err) =>
+                    anyhow!(err).context(format!("failed to load module {}", specifier)),
```
MarinPostma, line 372:
```suggestion
                    anyhow!(err).context(format!("failed to load module {specifier}")),
```

tsc_compile/src/lib.rs
```diff
@@ -365,7 +366,18 @@ impl Compiler {
...
+                ModuleGraphError::ResolutionError(err) => {
+                    let range = err.range().clone();
+                    anyhow!(err).context(format!("failed to resolve module (at {})", range))
```
MarinPostma, line 375:
```suggestion
                    anyhow!(err).context(format!("failed to resolve module (at {range})"))
```

MarinPostma (dismissed):
A few nits + not sure why you are making changes to the deno submodule here

honzasp:
> @honzasp is it necessary to bump the deno submodule? If not it might be preferable to do it in a separate PR.

I'm not sure, I didn't explicitly update it, it just changed automagically :/

tsc_compile/src/lib.rs
honzasp: I think I tried this but it did not work, though I don't remember why :)

espindola:
> > @honzasp is it necessary to bump the deno submodule? If not it might be preferable to do it in a separate PR.
> 
> I'm not sure, I didn't explicitly update it, it just changed automagically :/

It somehow moved to an older revision. Maybe you changed from an old branch and didn't update the submodule?

cli/tests/lit/bad-url.deno
```diff
@@ -13,6 +13,6 @@ EOF
 cd "$TEMPDIR"
 $CHISEL apply 2>&1 || true
 
-# CHECK: Error: parsing endpoints
+# CHECK: Error: could not compile endpoints
```
espindola, line 16: Please squash the test updates into the commits that cause the change. Having the tests pass on every revision is super helpful when bisecting.

server/src/deno.rs
```diff
@@ -175,13 +175,21 @@ impl deno_core::ModuleLoader for ModuleLoader {
...
         let handle = self.inner.lock().unwrap();
-        let code = handle.code_map.get(specifier).unwrap().clone();
-        let specifier = specifier.clone();
-        async move { wrap(&specifier, code) }.boxed_local()
+        if let Some(code) = handle.code_map.get(specifier).cloned() {
+            let specifier = specifier.clone();
+            async move { wrap(&specifier, code) }.boxed_local()
+        } else {
+            let err = anyhow!(
+                "chiseld cannot load module {} at runtime{}",
```
espindola, line 187: Do you have a test for this? We should only get here with bug in the CLI I think. Defensive programing is fine as we don't want to crash with a malicious user, but a comment is in order if that is the case.

tsc_compile/src/lib.rs
espindola: In what way this improves the error message? Is this something that deno_graph could do better? If so, could you open a PR with them?

espindola:
With just the first patch I get:

```
117 |             .with_context("could not bundle endpoints with esbuild (using node-style modules)");
    |              ------------ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected an `Fn<()>` closure, found `str`
    |              |
    |              required by a bound introduced by this call
```

tsc_compile/src/lib.rs
honzasp: Yes, the problem is that `ModuleGraphError` does not implement `Error::source()`, so we lose the causality chain from `err`. I also use the opportunity to make sure that the error message contains the URL of the module that failed to load.

honzasp:
It seems that trying to separate the PR into multiple commits added more problems that it fixed, I will just squash it all together.

server/src/deno.rs
honzasp: Yes, you can trigger this by `import { foo } from "http://foo.bar";` in Node mode (or just any other `http` or `https` import)

honzasp: I will add a lit test for this, is that ok?

tsc_compile/src/lib.rs
honzasp: The compiler complains:
```
error[E0277]: the trait bound `Result<std::string::String, deno_core::anyhow::Error>: From<Result<std::string::String, std::io::Error>>` is not satisfied
```

Apparently, `Result` does not implement `From` or `Into`. Why did we both think it does? :thinking:

server/src/deno.rs
espindola: Yes, a test is perfect. It does sound like a bug in the cli, but that can be fixed in a followup.

espindola (approved):

cli/src/cmd/apply/node.rs
```diff
@@ -141,8 +142,8 @@ pub(crate) async fn apply(
     Ok((endpoints_req, index_candidates_req))
 }
 
-fn npx<A: AsRef<OsStr>, C: AsRef<OsStr>>(
-    command: C,
+fn npx<A: AsRef<OsStr>>(
+    command: &'static str,
```
dekimir, line 146: Hehe, when I tried this, @glommer complained: https://github.com/chiselstrike/chiselstrike/pull/709#discussion_r793184089

honzasp: Using `&'static str` makes it easier to pass `command` to the closure for `spawn_blocking()`, so that we can print the failed subcommand. This is just a private helper function, so I took the liberty :)

cli/tests/lit/bad-url.deno
honzasp: Yes, I squashed all changes into a single commit.

tsc_compile/src/lib.rs
honzasp: Here we go :) denoland/deno_graph#164

cli/tests/lit/http-import.node
```diff
@@ -0,0 +1,17 @@
+# SPDX-FileCopyrightText: © 2021 ChiselStrike <info@chiselstrike.com>
```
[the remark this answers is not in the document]
honzasp: I opened issue #1572 for this.

PR #1558, opened 2022-07-13
Query helper

BearLemma:
First little simplification and quality life improvement for our endpoints.
Introduces Query structure to simplify parsing and working with query string.

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -650,8 +650,11 @@ export class ChiselCursor<T> {
  * @property {string} endpoint - The current endpoint being called.
  * @property {string} pathParams - This is essentially the URL's path, but with everything before the endpoint name removed.
  * @property {AuthUser} user - The currently logged in user. `undefined` if there isn't one.
+ * @property {Query} query - Helper structure containing parsed query string from the URL.
  */
 export class ChiselRequest extends Request {
+    public query: Query;
+
     constructor(
         input: string,
         init: RequestInit,
@@ -661,6 +664,7 @@ export class ChiselRequest extends Request {
         public user?: AuthUser | undefined,
     ) {
         super(input, init);
+        this.query = new Query(new URL(this.url).searchParams);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


MarinPostma (commented):
I there a way to make `get` a generic method?

api/src/chisel.ts
```diff
@@ -678,6 +682,57 @@ export class ChiselRequest extends Request {
     }
 }
 
+/** Query is a helper class used to access query parameters parsed from the URL.
+ */
+export class Query {
```
honzasp, line 687: Maybe we should also implement other methods from `URLSearchParams`:

- `entries()`, `keys()`, `values()` and `forEach()` for iteration
- `has()` for existence checking
- `getAll()` to get all occurences of one key (`get()` returns only the first value)

All these methods should directly call the corresponding methods of `params`.

honzasp: Or perhaps we can directly inherit from `URLSearchParams` and get all these methods for free? :thinking:

honzasp (commented):
I think this is great, but the API might be improved:
- Add more methods from `URLSearchParams`, such as `getAll()` or `entries()`
- Don't use the explicit default values in the getters, let the caller use `?? default` if they want to (this also makes `get()` compatible with `URLSearchParams`)
- Change the semantics of boolean parsing; at the very least, `"0"` should be treated as false.
(More details in comments)

api/src/chisel.ts
```diff
@@ -678,6 +682,57 @@ export class ChiselRequest extends Request {
...
+     * parameter exists, returns `undefined` or `defaultValue` if provided.
+     * If `paramName` key exists, the value is converted to lowercase and then compared to `'true'`.
+     * This means that the function returns false for anything other than variations of `true` string.
+     * @param paramName query parameter to be retrieved from the URL's query string.
+     * @param defaultValue default value to be returned when `paramKey` is not present.
+     */
+    getBool(paramName: string, defaultValue?: boolean): boolean | undefined {
+        const v = this.get(paramName);
+        if (v !== undefined) {
+            return v.toLowerCase() === "true";
```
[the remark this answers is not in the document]
honzasp: At the very least, we must accept `"1"` as a true value.

api/src/chisel.ts
honzasp: You are right, let's just implement these methods without inheriting from `URLSearchParams`.

honzasp (dismissed):

honzasp (approved):

PR #1561, opened 2022-07-14
Use our forked version of lit in chiselc

BearLemma:
Uses our forked version of lit crate in chiselc, adds convenience `config.always_show_stdout = false;` line.

### DIFF
--- a/chiselc/tests/integration_test.rs
+++ b/chiselc/tests/integration_test.rs
@@ -25,6 +25,7 @@ mod tests {
             config.add_extension("lit");
             config.constants.insert("chiselc".to_owned(), chiselc());
             config.truncate_output_context_to_number_of_lines = Some(80);
+            config.always_show_stdout = false;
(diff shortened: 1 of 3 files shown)


chiselc/tests/integration_test.rs
```diff
@@ -25,6 +25,7 @@ mod tests {
             config.add_extension("lit");
             config.constants.insert("chiselc".to_owned(), chiselc());
             config.truncate_output_context_to_number_of_lines = Some(80);
+            config.always_show_stdout = false;
```
espindola, line 28: LGTM, but why do we want to suppress stdout?

espindola (approved):

MarinPostma:
@BearLemma since we are using the patch in different location, what about creating a workspace level patch for lit? https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

PR #1565, opened 2022-07-14
Use an atomic instead of rayon::current_thread_index()

espindola:
This is an experiment to see if cures the "Address already in use"
errors.


MarinPostma (approved):

penberg:
@MarinPostma I don't think this pull request fixes the problem. I have had some success with turning on reuseport for Tonic (https://github.com/chiselstrike/chiselstrike/pull/1567), but it needs some more testing to really tell.

PR #1567, opened 2022-07-14
server: Enable SO_REUSEPORT for the Tonic server

penberg:
We're seeing the following error that results in `chisel wait` timing
out in CI runs:

```
Error: transport error

 Caused by:
     0: error creating server listener: Address already in use (os error 98)
     1: Address already in use (os error 98)
```

One possible explanation for this bug is that as we recycle port
numbers, we attempt to reuse a port that's still in TIME_WAIT state
(process terminated, but TCP stack not ready to give it back).

Therefore, like we already do for the API server, let's enable
SO_REUSEPORT for the Tonic server too.

### DIFF
--- a/server/src/rpc.rs
+++ b/server/src/rpc.rs
@@ -27,13 +27,18 @@ use chisel::{
 use deno_core::futures;
 use deno_core::url::Url;
 use futures::FutureExt;
+use hyper::service::Service;
+use hyper::service::{make_service_fn, service_fn};
+use hyper::Server;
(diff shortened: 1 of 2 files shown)


glommer:
reuseport won't do anything here.
It only works within the same process. We use it for the api server because we want various threads binding to the same port.

across process it has no effect (reuseaddr works across process)

penberg:
As @glommer points out, `SO_REUSEPORT` is bogus because it doesn't work across processes (`SO_REUSEADDR`) does.

PR #1584, opened 2022-07-20
make sure paths are respected within .gen

### LINKED ISSUE #1511: nested path alias
Consider following file structure..

image

Following import pathes are actually working.. but VSCode cannot find the files..

image

Otherwise relative as expected pathes are recognized by VSCode.. but cannot be found by chisel..

image

image

glommer:
There's currently a pretty serious problem with chiselc and endpoints
within subdirectories. The algorithm to figure out the name of the
generated file is to just take the filename() component of the endpoint
path, which has two big problems:

1. endpoints/foo/bar.ts will override endpoints/bar.ts in .gen.
2. import paths containing "../../" will not work.

The best way forward, especially because of relative paths for imports,
is to keep path structures the same, just replacing endpoints with .gen.

We'll have issues if the user uses either absolute paths or ./ as paths
in the manifest, but those are likely bad ideas (with the first
potentially being a security issue in the managed deployments) anyway.
So this patch also prevents them by checking early and making sure those
things are not present.

Fixes #1511

### DIFF
--- a/cli/src/cmd/apply/node.rs
+++ b/cli/src/cmd/apply/node.rs
@@ -36,10 +36,28 @@ pub(crate) async fn apply(
     let gen_dir = cwd.join(".gen");
     fs::create_dir_all(&gen_dir)?;
 
-    let chiselc_futures = endpoints.iter().map(|endpoint| {
+    let mut chiselc_futures = vec![];
+    for endpoint in endpoints.iter() {
         if optimize {
             let endpoint_file_path = endpoint.clone();
-            let gen_file_path = gen_dir.join(endpoint_file_path.file_name().unwrap());
+            let mut components = endpoint_file_path.components();
+            components.next();
+            let endpoint_rel_path = components.as_path();
+            anyhow::ensure!(
+                endpoint_file_path.is_relative(),
+                "malformed endpoint name {}. Shouldn't have reached this far",
+                endpoint_file_path.display()
+            );
+
+            let gen_file_path = gen_dir.join(&endpoint_rel_path);
+            let base = gen_file_path.parent().ok_or_else(|| {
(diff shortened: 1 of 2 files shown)


glommer:
@BearLemma can you take a look ?

cli/src/cmd/apply/node.rs
```diff
@@ -36,10 +36,28 @@ pub(crate) async fn apply(
...
+    let mut chiselc_futures = vec![];
+    for endpoint in endpoints.iter() {
         if optimize {
             let endpoint_file_path = endpoint.clone();
-            let gen_file_path = gen_dir.join(endpoint_file_path.file_name().unwrap());
+            let mut components = endpoint_file_path.components();
+            components.next();
+            let endpoint_rel_path = components.as_path();
+            anyhow::ensure!(
+                endpoint_file_path.is_relative(),
```
dekimir, line 47: Isn't it better to do this before calculating `endpoint_rel_path`?

cli/src/project.rs
```diff
@@ -111,8 +111,29 @@ impl Manifest {
...
 
     fn dirs_to_paths(dirs: &[String]) -> anyhow::Result<Vec<PathBuf>> {
+        // sucks to do this for all invocations but keeps things simple
+        let me = Path::new("./").canonicalize()?;
         let mut paths = vec![];
         for dir in dirs {
+            let p = Path::new(dir);
+            anyhow::ensure!(
+                p.is_relative(),
+                "{} is not relative to the current tree",
```
dekimir, line 121: Shouldn't we have a test for this?

dekimir (approved):
LGTM, but needs more tests.

Issue #1562, opened 2022-07-14
`chiselc` outputs wrong expression type for variables


penberg:
The following snippet in an endpoint:

```typescript
    const name = "Glauber";
     const filtered = Person.cursor().filter((p: Person) => {
       return p.name == name;
     }).select("age");

     const results = (await filtered.toArray()).map(p => JSON.stringify(p));
```

errors out with:

```
Unhandled exception: Error: Promise rejected without a handler: TypeError: Error parsing args at position 0: unknown variant `Identifier`, expected one of `Literal`, `Parameter`, `Property`, `Binary`
```

The issue is that `Identifier` is not something that the server actually understands. The proposed fix is to change `Literal` into `Value` and just use that for variables too.

PR #1585, opened 2022-07-20
Fix chiselc's output for lambdas using local variables.

BearLemma:
Closes https://github.com/chiselstrike/chiselstrike/issues/1562

It's really hacky though... Without types it's really hard to get it to be correct.


glommer (approved):

dekimir:
I see @glommer already approved.  Looks good, but I suggest a couple of follow-ups:
1. Add a test with actual code from #1562.  Maybe in `cli/tests/lit/filter-with-predicate.deno`.
2. Proofread `chiselc/docs/query-expr.md` and ensure it's still correct.  Right now it shows `chiselc` outputting `exprType: Identifier`, which I don't think we want.  (And it calls `person` a `post`, too!)

PR #1586, opened 2022-07-21
add tests for manifest rules

glommer:
We already had some, but not introduced more, rules about what paths
can be listed in the manifest.

Add tests for that. Also, it's overkill to use lit for this, when
rust unit tests would do.


glommer:
FYI @dekimir

cli/src/project.rs
```diff
@@ -302,3 +308,83 @@ pub(crate) fn project_exists(path: &Path) -> bool {
...
+endpoints = ["endpoints"]
+policies = ["policies"]
+"#,
+        );
+        let m = read_manifest_from(d.path()).unwrap();
+        m.models().unwrap();
+        m.policies().unwrap();
+        m.endpoints().unwrap();
+    }
+}
```
dekimir, line 390: Can we add an empty-string case?

cli/src/project.rs
```diff
@@ -302,3 +308,83 @@ pub(crate) fn project_exists(path: &Path) -> bool {
...
+        m.policies().unwrap();
+        m.endpoints().unwrap();
+    }
+
+    #[should_panic(expected = "has to be a subdirectory")]
+    #[test]
+    fn parse_curr_dir_fails() {
+        let d = gen_manifest(
+            r#"
+models = ["./"]
```
dekimir, line 364: Is the slash ending significant? What about just `.`?

dekimir (approved):

cli/src/project.rs
glommer: we can, but empty string is actually allowed (it is toml-equivalent to just not specifying it). It is legal not to specify either endpoints, models or policies

cli/src/project.rs
glommer: they should both work the same. I didn't feel the need to add a test for that.

PR #1590, opened 2022-07-26
Add documentation on how to deploy

penberg:


penberg:
@apoorv-on-git 👆

docusaurus/docs/Intro/deploying.md
```diff
@@ -0,0 +1,9 @@
+# Deploying
+
```
dekimir, line 2: We should explain here how deploying differs from running locally and why you may want to do it.

dekimir (approved):

PR #1591, opened 2022-07-26
Fix is_unique fields migration.

BearLemma:
Fixes unique field migration in the meta DB. It was likely a typo.


honzasp (commented):
Maybe we should have some test for this code? Ideally something that would have caught this typo.

honzasp (approved):

PR #1594, opened 2022-07-26
simplify hello endpoint

glommer:
Since chiselstrike 0.11, this code can be simpler.


MarinPostma (approved):

honzasp:
Note that this code does something else:
- when you `return new Response("hello")`, the response is `hello` with `content-type: text/plain`
- when you `return "hello"`, the response is serialized as JSON, so the body is `"hello"` with `content-type: application/json`

Not sure whether this was intentional?

honzasp:
@glommer @MarinPostma

glommer:
This is fine. 
This is just an example application, and the content was not set to text by design anyway.

PR #1597, opened 2022-07-27
remove rustfmt as build dependency in tonic_build

MarinPostma:
As I was trying to build chisel in a container, I realized that tonic_build had rustfmt feature enabled by default. This needlessly complicates the build process, and newer versions of tonic do not require that anymore.

For now, I have only removed the feature, but it may be desirable to bump tonic later on.


penberg:
@MarinPostma Looks good, but this is in draft mode.

Issue #1599, opened 2022-07-27
Default boolean values in model are not handled correctly


honzasp:
In this code, the field `isAwesome` has default value `true`:

```ts
import { ChiselEntity } from '@chiselstrike/api';

export class Human extends ChiselEntity {
    name: string;
    isAwesome: boolean = !true;
}
```

Expected behavior is either that the default value is `false`, or that `chisel apply` returns an error.

honzasp:
Note that if I try to do something similar with strings:
```ts
    nickname: string = "bo" + "b";
```
then I get an error from `chisel apply`: `error: expression not supported`. Same error should perhaps be reported in this case.

glommer:
After the fix in #1600 this code will be accepted.
However, we handle expressions differently than we handle literals, and those examples could be literals, but will end up being treated as expressions.

Leaving it open for a bit longer until we can treat them as the literals that they are.

honzasp:
Hmm, so what is the difference between a literal and expression default value? It might be confusing if `isAwesome: boolean = false` behaves differently from `isAwesome: boolean = !true`.

glommer:
it is confusing, which is why I am leaving this issue open and not closing with the other fix =p

The difference is that for literals, we can make the property hot-addable and/or removable.

PR #1600, opened 2022-07-27
allow arbitrary defaults

glommer:
This PR allow us to write things like:

  a: number = Date.now()

in defaults.

We need surprisingly little to make this work, which makes me very
happy. Essentially, Entity.build() will already execute those functions,
so it works just fine.

We just need to make sure that the schema evolution works, which also
happen to work just fine: for complex expressions, we just treat them
as "no defaults". That makes sense, because if you have something like
"Date.now()", the behavior the user expects for things that didn't
exist before is not to retrofill with the current date...

So we allow adding mandatory types with literals and change the error
messages accordingly. Complex expressions behave like no-default.

We do miss the opportunity to allow things like "string".length and "1 +
2", which are not literals, but are pure and could be executed compile
time. However this is good for now, and we could in the future expand
chiselc to allow for those expressions to be parsed as literal defaults.

### DIFF
--- a/cli/src/ts.rs
+++ b/cli/src/ts.rs
@@ -114,9 +114,12 @@ fn get_field_value(handler: &Handler, x: &Option<Box<Expr>>) -> Result<Option<(S
                     .ok_or_else(|| swc_err(handler, k, "unexpected empty expression"))?;
                 Ok(Some((format!("{}{}", op, value.0), value.1)))
             }
-            // Covers entity fields initialization.
-            Expr::New(_) => Ok(None),
-            x => Err(swc_err(handler, x, "expression not supported")),
+            // If the code is invalid, then parsing will reject this anyway. If it is valid
+            // but not a literal or unary, so we just behave as if there is no default as far
+            // as the type system is concerned. That means we cannot add this field to an existing
+            // schema (unless as optional), but in a new schema is fine. The runtime will execute
+            // this expression and end up with the correct default.
+            _ => Ok(None),
         },
     }
 }
@@ -324,6 +327,7 @@ fn parse_one_file<P: AsRef<Path>>(
                 // Right now just accept imports, but don't try to parse them.
                 // The compiler will error out if the imports are invalid.
             }
+            ModuleItem::Stmt(_) => {}
             z => {
                 handler.span_err(
                     z.span(),
--- a/cli/tests/lit/type-evolve.deno
+++ b/cli/tests/lit/type-evolve.deno
@@ -167,3 +167,87 @@ rm $TEMPDIR/endpoints/test_indexes.ts
 
 $CHISEL apply --allow-type-deletion
 # CHECK: Model defined: Evolving
+
+### get / set helpers for the tests below
+cat << EOF > "$TEMPDIR/endpoints/get.ts"
+import { Defaults } from "../models/defaults.ts";
+
+export default async function chisel(req: Request) {
+    return Defaults.findAll()
+}
+EOF
+
+cat << EOF > "$TEMPDIR/endpoints/set.ts"
+import { Defaults } from "../models/defaults.ts";
+
+export default async function chisel(req: Request) {
+    const x = Defaults.build();
+    await x.save();
+}
+EOF
+
+## Test able to use fields with complex defaults...
+cat << EOF > "$TEMPDIR/models/defaults.ts"
+import { ChiselEntity } from "@chiselstrike/api";
+
+export class Defaults extends ChiselEntity {
+    a: number  = "string".length
+}
+EOF
+
+$CHISEL apply
+# CHECK: Model defined: Defaults
+
+$CURL -d '{}' $CHISELD_HOST/dev/set
+# CHECK: HTTP/1.1 200 OK
+
+$CURL  $CHISELD_HOST/dev/get
+# CHECK: "a": 6
+
+## But not able to add other complex fields ...
+cat << EOF > "$TEMPDIR/models/defaults.ts"
+import { ChiselEntity } from "@chiselstrike/api";
+
+export class Defaults extends ChiselEntity {
+    a: number = "string".length
+    b: string = JSON.stringify({a: "test2"});
+}
+EOF
+
+$CHISEL apply 2>&1 || true
+# CHECK: Error: unsafe to replace type
+
+## ... unless they are optional!
+## Also check that defaults can be whatever, including other models
+cat << EOF > "$TEMPDIR/models/defaults.ts"
+import { ChiselEntity } from "@chiselstrike/api";
+
+export class Aux extends ChiselEntity {
+    aux: string = "aux"
+}
+
+function myFunction() : string {
+	return "funcreturn"
+}
+
+export class Defaults extends ChiselEntity {
+    a: number = "string".length
+    b?: string = JSON.stringify({a: "test2"});
+    c?: string = myFunction()
+    d?: Aux = new Aux()
+}
+EOF
+
+$CHISEL apply
+# CHECK: Model defined: Aux
+# CHECK: Model defined: Defaults
+
+$CURL -d '{}' $CHISELD_HOST/dev/set
+# CHECK: HTTP/1.1 200 OK
+
+$CURL  $CHISELD_HOST/dev/get
+# CHECK: "a": 6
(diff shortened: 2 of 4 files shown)


honzasp:
Hmm, maybe this fixes #1599 ?

glommer:
it does - as long as you don't expect those to behave as literal values (for that we'd need chiselc support)

cli/src/ts.rs
```diff
@@ -324,6 +327,7 @@ fn parse_one_file<P: AsRef<Path>>(
                 // Right now just accept imports, but don't try to parse them.
                 // The compiler will error out if the imports are invalid.
             }
+            ModuleItem::Stmt(_) => {}
```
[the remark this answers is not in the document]
glommer: allowing non-class statements in the file.
They will be ignored by the parser, meaning we won't extract chisel information from it for the purpose of Model creation, but they are allowed in the file now

Allow us to write things like this:
```
function foo(): number { <== statement, ignored
    return Math.random()
}

export class Jan extends ChiselEntity {
   a: number = foo();
}
```

Note that we still compile that file in the project, so parse errors in the stament are still caught, but we don't look at it when doing the model.

cli/tests/lit/type-evolve.deno
```diff
@@ -167,3 +167,84 @@ rm $TEMPDIR/endpoints/test_indexes.ts
...
+
+export class Defaults extends ChiselEntity {
+    a: number = "string".length
+    b?: string = JSON.stringify({a: "test2"});
+    c?: string = myFunction()
+}
+EOF
+
+$CHISEL apply
+# CHECK: Model defined: Defaults
```
dekimir, line 250: Please add some `set` and `get` invocations here.  (Including checking for correct calculation of the new fields on previously existing entities.)

cli/tests/lit/type-evolve.deno
```diff
@@ -167,3 +167,84 @@ rm $TEMPDIR/endpoints/test_indexes.ts
...
+import { ChiselEntity } from "@chiselstrike/api";
+
+function myFunction() : string {
+	return Math.random().toString()
+}
+
+export class Defaults extends ChiselEntity {
+    a: number = "string".length
+    b?: string = JSON.stringify({a: "test2"});
+    c?: string = myFunction()
```
dekimir, line 245: Please add a testcase for fields of entity type.

dekimir (changes requested):

cli/src/ts.rs
```diff
@@ -114,9 +114,12 @@ fn get_field_value(handler: &Handler, x: &Option<Box<Expr>>) -> Result<Option<(S
...
             }
-            // Covers entity fields initialization.
-            Expr::New(_) => Ok(None),
-            x => Err(swc_err(handler, x, "expression not supported")),
+            // If the code is invalid, then parsing will reject this anyway. If it is valid
+            // but not a literal or unary, so we just behave as if there is no default as far
+            // as the type system is concerned. That means we cannot add this field to an existing
+            // schema (unless as optional), but in a new schema is fine. The runtime will execute
+            // this expression and end up with the correct default.
+            _ => Ok(None),
```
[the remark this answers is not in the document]
dekimir: This basically introduces lenses through a side door.  I'm surprised that more work isn't required for lenses, even if we disallow auto-migration for mandatory fields.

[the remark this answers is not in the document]
glommer: what does this have to do with lenses?

cli/tests/lit/type-evolve.deno
glommer: What do you mean by "including checking for correct caltulation of the new fields on previously existing entities" ?
Together with the comment on lenses, I suspect you didn't fully read the commit description. Default fields that are not literals will not change existing fields.

cli/tests/lit/type-evolve.deno
glommer: This works when I apply on a project:
```
export class Aux extends ChiselEntity {
    x: string = "bar";
}

export class Defaults extends ChiselEntity {
    a: number = "string".length
    b?: string = JSON.stringify({a: "test2"});
    c?: string = myFunction()
    d?: Aux = new Aux()
}
```

It doesn't work in the test and I have no idea why - it's driving me crazy. updates soon (I hope)

cli/src/ts.rs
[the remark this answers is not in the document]
dekimir: It tells you how to calculate a newly added field, which allows creating the new model version from the old one.

cli/tests/lit/type-evolve.deno
dekimir: > What do you mean by "including checking for correct caltulation of the new fields on previously existing entities" ?

I mean invoke `/set` to create an entity with just field `a`, then extending model with fields `b` and `c`.  That one existing entity should then have `b` and `c` added correctly.

dekimir: Hm, maybe not.  Because `b` and `c` are optional...

glommer: that's not what this PR does, and what you're describing sounds like magic =p

Again, this is in the PR description: new objects will have that default. That means that those fields can only be added if they are optional (same evolution rules as before).

cli/src/ts.rs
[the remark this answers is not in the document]
glommer: this PR doesn't do this.

cli/tests/lit/type-evolve.deno
glommer: FYI, found the issue.

The issue is that if in the same apply, we add a new model M1, and then add a property to an existing entity that is of type M1, it fails on type lookup: we'll lookup type M1 but it doesn't exist yet. It's a pre-existing problem and I just managed to reproduce without this PR, so I will for now just open an issue and move on.

glommer: https://github.com/chiselstrike/chiselstrike/issues/1605

glommer: I added a test that tests that we can have non-literal defaults for entities, but they have to be there from the start due to the issue above.

I don't think it is a priority to get to it now.

glommer: lol, it's not a priority but I got annoyed and found the reason

server/src/datastore/meta/mod.rs
```diff
@@ -510,7 +510,35 @@ impl MetaService {
...
+                }
+                Err(_) => {
+                    failures.push(row);
+                }
+            }
+        }
+
+        // Retry once for failures. The reason we may want to retry is that if you have a model (A)
+        // that has another model (B) as a property, load_type_fields may fail, because B is not
+        // yet loaded.
```
dekimir, line 535: Wasn't this (or a variant of it) solved by 897dbc3d??

dekimir (changes requested):

server/src/datastore/meta/mod.rs
glommer: I explain that explicitly in the comment, although didn't mention the PR by number.
That only solves the case where you're adding all types at the same time.

We can't reuse that code easily either because it is implemented using the RPC messages with the textual description of the model. It's much easier to just retry load.

glommer:
@dekimir I updated this with your comments

dekimir (approved):

server/src/datastore/meta/mod.rs
dekimir: I suspect that retrying only once isn't enough, though I couldn't prove that with a simple example I ran.  But theoretically, if one failure depends on another failure (so neither is added to `ts` in the first `for` loop), then it is possible for the corresponding `load_type_fields()` in the second loop to error out, if failures are reprocessed in the wrong order.

glommer: cycles could lead to that, but they are currently disallowed.

dekimir: Cycles are a red herring, the failure chain only needs two links A > B > C for this to occur.  But it's still impossible, for another reason: one failure cannot chain on another.  A failure occurs only when an existing type adds a field of a new type.  All the new types are topo-sorted, so their field types always resolve on the first try.  Therefore, the chain stops right at the first link.

PR #1618, opened 2022-08-04
Refactor Deno code and implement new routing in chiseld

honzasp:
This PR introduces the new `RouteMap` API, moves routing to TypeScript, rewrites Deno interaction in `chiseld`
and refactors other parts of `chiseld` and related crates.

**Summary of changes:**

- Abandon the thread-per-core architecture: previously, there were N workers in different threads, each with
  one Deno runtime and each run code for all versions. Now, we still have one Deno runtime per worker thread
  (this is required by V8), but one runtime only handles a single version. See:

  - server/src/server.rs
  - server/src/trunk.rs
  - server/src/worker.ts
  - server/src/version.rs

- Simplify the Deno interop, so that we only call Rust from JavaScript (using `#[op]`-s), but we never call
  JavaScript from Rust. We also never attempt to reload code in a running Deno runtime, we simply start a new
  runtime when needed. See:

  - api/src/run.ts
  - api/src/serve.ts
  - server/src/ops/

- Move all routing functionality to TypeScript, as described in the routing RFC. See:

  - api/src/routing.ts
  - api/src/special.ts (Swagger introspection)

- Refactor `TypeSystem` and policies to make versions independent. See:

  - server/src/types/
  - server/src/policies.rs

- Introduce a strong tree hierarchy of Tokio tasks: every task should be `.await`-ed by another task. This
  makes the control flow easier to reason about and ensures that errors or panics are not swallowed. (This is
  not my idea, google "structured concurrency" or "go statement considered harmful"). See:

  - server/src/server.rs
  - server/src/api.rs
  - server/src/rpc.rs
  - server/src/trunk.rs

- Use the term "version id" to refer to the string identifier of a version (such as "dev" or "main"). In a
  future PR, the term "version" will be replaced with "branch".

- Don't try to guess absolute URLs for CRUD paging, but use relative URLs like everybody else. Guessing the
  schema, host and port is not robust: for example, the `Host` header gives us information about the host and
  port, but we don't know whether the client uses `http` or `https`.

- Add the file-based routing functionality into `chisel apply`. See:

  - cli/src/routes.rs
  - cli/src/apply/

- Adapt integration tests to the changes.


honzasp:
This PR is now ready for review. Pekka and Jenda, may I ask both of you to do the review? This is a big
change, so we need as many eyeballs as we can get to find any potential issues.

cc @penberg @BearLemma 

Please, **don't try to review the PR by scrolling through the Github diff**. You will be much happier if you
simply clone the branch into a temporary directory and use your IDE to read the code. Your IDE is much better
interface for reading and understanding code than Github!

I suggest that you start by reading PR description and studying the referenced files. When you gain some basic
understanding how things work, then you can use the Github diff as a guide to walk through the diff, switching
to you IDE whenever you encounter a bigger change.

Don't hesistate to write comments when something is unclear, even though it is correct. I would like to make
sure that the code is understandable, and I'm happy to add comments or refactor the code to make it more
clear.

This PR is still a draft, because I will add these changes:

1. Update the Docusaurus documentation to reflect the API changes.
2. Add integration tests that test the new functionality. I will also probably modify the new test framework
   using this experience.

But I think we should start the review now and not wait for these finishing touches.

Lastly, I'm open to applying parts of this PR separately, like we did in #1635: don't modifying this branch,
but applying the changes to `main` separately (so that they disappear from the diff of this PR).

*This is a big PR, but I'm confident that we can make this :)*

penberg:
@honzasp We use `pub(crate)` wherever possible _on purpose_ since day one because it allows us to detect unused code. Why are you switching over to `pub` everywhere?

penberg:
@honzasp Why is the `endpoints` directory renamed to `routes` in the tests? Are users also expected to do that for their projects?

penberg:
@honzasp The policy on copyright years is that we never update it on existing files.

honzasp:
> @honzasp We use `pub(crate)` wherever possible _on purpose_ since day one because it allows us to detect unused code. Why are you switching over to `pub` everywhere?

Dead code detection in Rust is smart: if you make an item `pub`, but don't export it from the crate and don't use it inside of the crate, it will be reported as dead code. So we can simply mark everyting `pub`, and use `pub(crate)` only at the top level of the crate, in `server/src/lib.rs`. The result is the same, but with less lexical noise.

honzasp:
> @honzasp Why is the `endpoints` directory renamed to `routes` in the tests? Are users also expected to do that for their projects?

Because the concept of an "endpoint" disappeared, so there is no reason to keep it in the directory names. Existing users don't have to do anything, because their `Chisel.toml` points to the `endpoints/` directory (and `endpoints` in `Chisel.toml` is now an alias for `routes`).

honzasp:
> @honzasp The policy on copyright years is that we never update it on existing files.

Oh, I'm an idiot, I should have checked this with you before changing this everywhere :( Does this policy also apply to files that were almost completely rewritten, like `server/src/server.rs`?

penberg:
> @honzasp Why is the `endpoints` directory renamed to `routes` in the tests? Are users also expected to do that for their projects?

To answer my own question: there's backward compatibility because most projects will have `Chisel.toml` with `endpoint` setting that `chiseld` knows to read in absence of `routes`.

However, I am not entirely convinced `routes` is the right name going forward. With the streaming support I am adding, there will be a `events` directory. In the future, I am hoping we will add support for periodic batch jobs in a `jobs` folder. So I kind of like `endpoints` because it hints towards "REST endpoints".

@honzasp what's your rationale for changing the directory name?

honzasp:
> @honzasp what's your rationale for changing the directory name?

The rationale is that the concept of an "endpoint" does not exist anymore and the structure that is defined by the files in that directory creates a "route map", whose entries are "routes". Using the name "endpoint" in just the directory name and nowhere else is weird.

The rationale for _that_ is that using a single name for one concept reduces confusion, which makes our users happier and more productive when using our services :)

(By the same reasoning, we should rename the "models" directory to "entities", because it contains definitions of entities and we don't use the name "model" anywhere else.)

api/src/crud.ts
```diff
@@ -1,182 +1,168 @@
...
-    POST: CRUDMethodSignature<T, E, P>;
-    PUT: CRUDMethodSignature<T, E, P>;
-    PATCH: CRUDMethodSignature<T, E, P>;
-    DELETE: CRUDMethodSignature<T, E, P>;
-};
+>(
+    entity: E,
+    config?: {
+        createResponse?: CRUDCreateResponse;
+        getAll?: boolean;
```
barbieri, line 51: would you mind still keeping the behavior where the code could be customized? Like a function | boolean, this way any pre-post processing is easily done, the code here wouldn't become much more complex.

api/src/crud.ts
```diff
@@ -1,182 +1,168 @@
...
+        patch?: boolean;
+        deleteAll?: boolean;
+        deleteOne?: boolean;
+    },
+): RouteMap {
+    const createResponse = config?.createResponse ?? responseFromJson;
+    const routeMap = new RouteMap();
+
+    // Returns all entities matching the filter in the `filter` URL parameter.
+    async function getAll(req: ChiselRequest): Promise<Response> {
```
barbieri, line 64: this function is huge, would you mind moving these all to toplevel functions that takes `entity` as first parameter? This way you can easily just bind the generic function to the current entity, if needed (ie: inside the `if` branch below).

api/src/crud.ts
```diff
@@ -1,182 +1,168 @@
...
+        if (!orig) {
+            return createResponse(
+                "object does not exist, cannot PATCH",
+                404,
+            );
+        }
+        mergeDeep(
+            orig as unknown as Record<string, unknown>,
+            await req.json(),
+        );
```
barbieri, line 122: beaware this could modify the `id` as well, which would cause problems. I think it could also cause some issues with extra stuff being absorbed, when it shouldn't... just like there is `entity.build()`, maybe there should be be an entity.merge? or call `entity.build(await req.json())` in here?

api/src/crud.ts
```diff
@@ -1,182 +1,168 @@
...
+    }
+    if (config?.patch ?? true) {
+        routeMap.patch("/:id", patch);
+    }
+
+    // Deletes all entities matching the filter in the `filter` URL parameter.
+    async function deleteAll(req: ChiselRequest): Promise<Response> {
+        await deleteEntitiesCrud(entity, Array.from(req.query));
+        return createResponse(
+            `Deleted entities matching ${new URL(req.url).search}`,
```
barbieri, line 134: why not use the same `req.query`?

api/src/crud.ts
```diff
@@ -185,240 +171,14 @@ async function fetchEntitiesCrud<T extends ChiselEntity>(
...
         {
             typeName: type.name,
-            url,
+            urlQuery,
         },
         requestContext,
     );
 }
-
-const defaultCrudMethods: CRUDMethods<ChiselEntity, GenericChiselEntityClass> =
```
barbieri, line 200: in other words, I'd keep something like this one for the crud operations above.

api/src/crud.ts
barbieri: or to make code simpler, it could be `function | null`, this way null is handled as false/remove the method, you can spread the standard functions and you get everything for free (as the old code below):
 - if pass a custom function, it's used (overrides the spread)
 - if pass null, method is disabled (overrides the spread)
 - if nothing is passed, uses the default CRUD functions, declared as constant)

api/src/http.ts
```diff
@@ -0,0 +1,150 @@
...
+import { ChiselRequest } from "./request.ts";
+import { Handler, Middleware, Router, RouterMatch } from "./routing.ts";
+import { opAsync, opSync, responseFromJson } from "./utils.ts";
+
+// HTTP request that we receive from Rust
+export type HttpRequest = {
+    method: string;
+    uri: string;
+    headers: [string, string][];
+    body: Uint8Array;
```
barbieri, line 13: add `| undefined` here? Or at least declare 2 that would imply in the correct type, ex:

```ts
type HttpRequestCommon = {
    uri: string;
    headers: [string, string][];
    routingPath: string;
    userId: string | undefined;
};

type HttpRequestWithBody = HttpRequestCommon & {
    method: 'PUT' | 'PATCH';
    body: Uint8Array;
};

type HttpRequestWithoutBody = HttpRequestCommon & {
  method: 'GET' | 'HEAD' | 'DELETE';
};

type HttpRequest = HttpRequestWithBody | HttpRequestWithoutBody;
```

api/src/http.ts
```diff
@@ -0,0 +1,150 @@
...
+
+        // read the response body before committing the transaction, because user
+        // code might still be running while the response is streaming
+        responseBody = await response.arrayBuffer();
+
+        await opAsync("op_chisel_commit_transaction");
+    } catch (e) {
+        let description = "";
+        if (e instanceof Error && e.stack !== undefined) {
+            description = e.stack;
```
barbieri, line 88: this is usually bad in production

barbieri: nevermind, just goes to log... there is a TODO below.

api/src/http.ts
```diff
@@ -0,0 +1,150 @@
...
+        } catch (e) {
+            console.error(`Error when rolling back transaction: ${e}`);
+        }
+
+        // return an empty response to avoid leaking details about the user error
+        // TODO: perhaps we should introduce a "debug mode" that would return a nice error response?
+        return emptyResponse(500);
+    }
+
+    return {
```
barbieri, line 107: move this return inside `try` and make all of the variables `const`, so you avoid mistakes by any kind and code is shorter.

api/src/http.ts
```diff
@@ -0,0 +1,150 @@
...
+): Promise<Response> {
+    return handleMiddlewareChain(
+        routerMatch.middlewares,
+        routerMatch.handler,
+        request,
+    );
+}
+
+async function handleMiddlewareChain(
+    middlewares: Middleware[],
```
barbieri, line 126: I'd change the signature to contain one more parameter `currentMiddlewareIndex` and then you can avoid creating/destroying the array on every step.

Your `if (middlewares.length == 0) {`  becomes `if (currentMiddlewareIndex === middlewares.length) {` and so on.

api/src/http.ts
```diff
@@ -0,0 +1,150 @@
...
+    } else {
+        // call the middleware handler, passing a callback that will continue in the middleware chain
+        const next = (request: ChiselRequest) =>
+            handleMiddlewareChain(middlewares.slice(1), handler, request);
+        return middlewares[0].handler.call(undefined, request, next);
+    }
+}
+
+function emptyResponse(status: number): HttpResponse {
+    return { status, headers: [], body: new Uint8Array(0) };
```
barbieri, line 149: move to constant?

barbieri:
```ts
const emptyResponseBase = { headers: [], body: new Uint8Array(0) };

function emptyResponse(status: number): HttpResponse {
  return { status, ... emptyResponseBase };
}
```
or just propagate the shortcut to return a number and pass it along handling directly since it's common

api/src/kafka.ts
```diff
@@ -0,0 +1,87 @@
...
+import { opAsync, opSync } from "./utils.ts";
+
+export type KafkaEvent = {
+    topic: string;
+    key: Uint8Array;
+    value: Uint8Array;
+};
+
+export class TopicMap {
+    topics: Record<string, EventHandler>;
```
barbieri, line 13: given that the event encoding is not always well defined, there is protobuf/avro/... I'd promote a `ParseEvent` to first class here, sibling of `EventHandler`, you can provide UInt8Array directly to it and offer a pre-defined `fromJSON` version, making such case simpler as it doesn't require a schema (unlike protobuf/avro/...)

barbieri: ie:
```ts
type ChiselEvent<K, V> = {
  key: K;
  value: V;
};
type ParseEvent<K, V> = (rawKey: UInt8Array, rawKey: UInt8Array) => ChiselEvent<K, V>;
type HandleEvent<K, V> = (event: ChiselEvent<K, V>) => Promise<void>;
type EventHandler<K, V> = {
  parser: ParseEvent<K, V>;
  handler: HandleEvent<K, V>;
};
```

and there is no need to have a `class` since it's better done as an actual JS object declared `as const`, so you get strict type checking, a given topic -> parser/handler.

api/src/kafka.ts
```diff
@@ -0,0 +1,87 @@
...
+            );
+        }
+        return buffer;
+    }
+
+    // create the `ChiselEvent` object
+    const chiselEvent = {
+        key: toBuffer(event.key),
+        value: toBuffer(event.value),
+    };
```
barbieri, line 64: then here would call `const chiselEvent  = topicMap[topic].parser(event.key, event.value)`

api/src/request.ts
```diff
@@ -4,39 +4,34 @@ import type { AuthUser } from "./datastore.ts";
...
+ * @property {string} versionId - The current API Version.
  * @property {AuthUser} user - The currently logged in user. `undefined` if there isn't one.
- * @property {Query} query - Helper structure containing parsed query string from the URL.
+ * @property {Query} query - Helper class containing parsed query string from the URL.
+ * @property {Params} params - Helper class containing parameters from the URL path.
  */
 export class ChiselRequest extends Request {
+    public path: string;
+    public versionId: string;
+    public user: AuthUser | undefined;
```
barbieri, line 16: readonly fields? or at least flag the whole ChiselRequest as `Readonly<T>` before passing to the handlers

api/src/routing.ts
```diff
@@ -0,0 +1,289 @@
...
+    req: ChiselRequest,
+) => ResponseLike | Promise<ResponseLike>;
+
+/** Anything that we can convert to a `Response`:
+ *
+ * - `Response` is not modified in any way
+ * - `string` is converted using `new Response(string)`
+ * - Other values are converted to JSON using `responseFromJson()`
+ */
+export type ResponseLike = Response | string | unknown;
```
barbieri, line 186: like said, you could `| number` here and handle it as such.

api/src/crud.ts
honzasp: The idea was that if you want to customize a route, you will simply define it yourself. For example, if you wanted to customize the `PUT /:id` route, you would say something like:

```
crud(Entity, {put: false}) // don't generate the default `PUT /:id` route...
    .put('/:id', async function (req) { ... }) // but define it yourself
```

Do you think this makes sense?

The problem with this approach is that you as the user don't have access to the builtin CRUD functionality in `fetchEntitiesCrud` and `deleteEntitiesCrud`. We can resolve this in multiple ways, for example by simply `export`-ing these functions, but I wanted to start simple and add the functionality gradually.

api/src/crud.ts
honzasp: I decided to keep the handler functions like `getAll` as closures in the `crud()` function, because they all use `entity` and `createResponse` from the scope of `crud()`. If they were hoisted to the top level, the code would look something like:

```typescript
async function getOne<
    T extends ChiselEntity,
    E extends ChiselEntityClass<T>,
>(
    entity: E,
    createResponse: CRUDCreateResponse,
    req: ChiselRequest,
): Promise<Response> {
    ...
}

// and inside crud()
routeMap.get("/:id", (req) => getOne(entity, createResponse, req))
```

I'm not sure whether this would be better than the current code; what would be the reasons for that? Or could we perhaps achieve that objective in more compact way?

api/src/crud.ts
honzasp: Yeah, I completely agree that this is very shady, I simply copied this code from the previous CRUD implementation. Note that `entity.build()` also calls `mergeDeep()` in exactly the same way! The problem is that we have no idea what are the properties of `entity` and their types. I'm not sure whether there is an easy solution, other than a major redesign of the data model API.

api/src/crud.ts
honzasp: Ah, good catch! I'll replace it with `req.query.toString()` :)

api/src/http.ts
honzasp: Hmm, what would be the motivation for this? When the request has no body, `HttpRequest.body` is simply empty. (Technically, an HTTP request always has a body, even though it should be empty for some methods.)

(Note that this type is passed to JavaScript from Rust, so this type is the JS representation of `HttpRequest` from `server/src/http.rs`.)

api/src/http.ts
honzasp: Ok, why not; there used to be more code between the `try` statement and the `return` statement, but the structure no longer makes sense when this code is removed.

api/src/http.ts
honzasp: Ok; I wrote the code in recursive Haskell-ish style in an attempt to make it as clear as possible, but your version with explicit index is probably better :)

api/src/http.ts
honzasp: What would be the reason to use `{ status: 500, ... emptyResposeBase }` instead of `emptyResponse(500)`? Are you concerned with the overhead of the function call and the two allocations?

api/src/kafka.ts
honzasp: The event handling was introduced in Pekka's PR #1634, in this PR I just want to reimplement the same functionality without modifying it or introducing new features (this PR is already huge!).

However, I agree that the current support for Kafka event consuming is very rudimentary (it is just a proof of concept, anyway). We should definitely provide some way to at least parse the `ArrayBuffer` as JSON, though I would prefer to simply provide a function `(ArrayBuffer) -> JSONValue` rather than coupling the parsing of bytes to the dispatch of events.

I'm not sure I understand your proposed API: do you want the event handlers in `events/` to export the `EventHandler<>` type, something like this?

```typescript
export default {
    parser: eventJsonParser,
    handler: async function (event: ChiselEvent<JSONValue, JSONValue>) {
        ...
    },
}
```

honzasp: The reason why I use a `TopicMap` class with a single property of type `Record<string, EventHandler>`, instead of using `Record<string, EventHandler>` directly, is to maintain the close analogy with `RouteMap`.

api/src/request.ts
honzasp: Ah, this is a good idea :) But we may keep the option for the user to modify the `ChiselRequest`, for example in middlewares. Maybe we should ask Jenda (@BearLemma) for his opinion?

api/src/routing.ts
honzasp: You mean to replace the use of `emptyResponse()` in `handleHttpRequest()` in `http.ts`? Note that `handleHttpRequest` returns a `HttpResponse` which is passed directly to Rust.

The `ResponseLike` type is used as a return type of the user-defined handler. We _do_ allow to return a number, but the semantics is that we call `responseFromJson` to convert it to a JSON response with status 200.

api/src/request.ts
honzasp: We decided that there is no good reason for the user to modify these properties, so `readonly` be it :) (Also, we can always lift the restriction later.)

honzasp:
Thanks Gustavo for your review, I applied your suggestions in the latest commit :) I left replies where I would like some clarifications, so we can continue the discussion in the threads.
cc @barbieri

api/src/crud.ts
dekimir: Whatever you decide, please also update the [documentation](https://github.com/chiselstrike/chiselstrike/blob/791fea8d22be8912f8ad788820cf312bba8fd38d/docusaurus/docs/Intro/routes.md?plain=1#L18).

barbieri: @honzasp ok, makes sense... I think the other way is more "standard" as you get the paths and all defined, just need to do some extra logic/validation.

But your approach also has some advantages, since you don't get the paths and all, you can do whatever you want, that is more flexible.

api/src/crud.ts
barbieri: your code would be smaller, did grew too big. You can get the "binding" either as you wrote or by using `fn.bind()` (single line)

api/src/crud.ts
barbieri: yeah, I saw the old ("suspicious") code, but since much is changing, I'd fix it as well at least by using `build()`.

If `build()` is broken we can fix it later, but we should be doing a safe approach in there, either going back to the Rust part where it knows the properties + types, either by generating some json schema and using Ajv.validate with coerce/removeAdditionalProperties (and then you can also generate the coreapi/swagger)...

TypeORM will use TypeScript annotations and metadata to get the columns (they force us to `@Column`), but maybe going back to Rust you can get the AST the same way you control what goes to the DB.

So `mergeDeep` per see is okay, the issue is not checking what was received in `req.json()`.

api/src/http.ts
barbieri: with the approach I said you can force TS to hint you, if you:

```ts
if (req.method === 'GET') {
  req.body // errors, since it doesn't exist
}
if (req.method === 'POST') {
  req.body // is always a string
}
```

if you want to force it to be `undefined`, just change the `HttpRequestWithoutBody` to include `body: undefined`. This way, if you ever use `req.body` the TSC will warn you "this may be undefined".

If you ALWAYS use `body: string | undefined` then in methods you KNOW there will be a body, you still need to check it... which is an usability issue.

barbieri: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions

api/src/http.ts
barbieri: yeah, you don't need to keep doing the allocations... but my personal preference is to handle `number` just like you handle `string` as a first class citizen, then it can go smooth, including user's code :-)

api/src/kafka.ts
barbieri: yes, you'd export that, BUT you should be linking the types in a way... it's better if you can get TSC to help instead of `JSONValue`.

For instance, if we can use ajv + JSONSchema you can even infer the typescript type from it, see https://www.fastify.io/docs/latest/Reference/TypeScript/#fastify-type-providers or even our own https://github.com/profusion/json-schema-to-typescript-definitions/

this way you get the type expected in that message.

As for `RouteMap`, yeah... I didn't comment on that one, but definitely that one could as well be a single object mapping stuff that is passed along. The actual "compilation" of stuff can happen as user-site, with the whole tree at hand. Ie you keep passing `type RouteMap = { [path: string]: Handler | RouteMap }`, then when you'll create your dispatcher you build the hierarchy... for example you could concatenate all the paths and use an object or `Map`, or you could even split the path and walk parts. In other words, it would be much simpler to use (less `map.route(xxx)`) and possibly more efficient so you're getting the whole tree at once at the end and can process the info in one go.

api/src/http.ts
honzasp: But I don't think we would get any type safety that way, because we don't use the `HttpRequest` type too much, we simply convert it to `ChiselRequest`. In fact, the `body` property is always here even for GET and HEAD, it just happens to be empty most of the time.

api/src/routing.ts
honzasp: I understand the point here, but it would be quite massive breaking change. Folks, what do you think? Should `return 404` return an empty response with status 404 instead of a JSON response saying `404`? @BearLemma @penberg @dekimir

api/src/crud.ts
honzasp: So what you are proposing is to do here is the following:
```
const orig = await entity.findOne(...);
mergeDeep(orig, await req.json());
const updated = entity.build(orig);
await updated.save();
```
Is that correct?

api/src/http.ts
```diff
@@ -0,0 +1,171 @@
...
+export async function handleHttpRequest(
+    router: Router,
+    httpRequest: HttpRequest,
+): Promise<HttpResponse> {
+    const routerMatch = router.lookup(
+        httpRequest.method,
+        httpRequest.routingPath,
+    );
+    if (routerMatch === "not_found") {
+        return emptyResponse(404);
```
CodingDoug, line 37: I've noticed that throughout this file (and possibly elsewhere) there are a bunch of hard-coded HTTP codes that could probably be replaced by symbols that are more readable and less prone to typos.

dekimir (changes requested):

api/src/routing.ts
dekimir: I think `return 404` is much more confusing and complicated to explain than `return emptyResponse(404)`.  The convenience is not worth the complexity.

api/src/crud.ts
```diff
@@ -1,182 +1,173 @@
...
-    params: P,
-    url: URL,
-    createResponse: CRUDCreateResponse,
-) => Promise<Response>;
-
 /**
- * A dictionary mapping HTTP methods into corresponding REST methods that process a Request and return a Response.
+ * Generates a route map to handle REST methods GET/PUT/POST/DELETE for this entity.
+ * @example
+ * Put this in the file 'endpoints/comments.ts':
```
dekimir, line 23: `endpoints` or `routes`?

api/src/crud.ts
dekimir:
```
crud(Entity, {put: false}) // don't generate the default `PUT /:id` route...
    .put('/:id', async function (req) { ... }) // but define it yourself
```
Why force the user to say `put` twice?  Can we have `crud(Entity).put('/:id/', ...)`?  The very existence of `.put()` means a custom PUT route replaces the default CRUD one.

api/src/crud.ts
dekimir: I think this is fine as is for now.  @barbieri if you feel like it, prepare a follow up that breaks the function down into parts, and we'll consider it separately.

api/src/crud.ts
dekimir: Please don't add any more "fixes" to this huge PR.  Reusing existing code is preferable; we'll fix it later.

api/src/routing.ts
```diff
@@ -0,0 +1,310 @@
...
+
+export type RouteMapLike =
+    | RouteMap
+    | Record<string, Handler>
```
dekimir, line 207: Please explain what the `string` is.

api/src/http.ts
honzasp: Good point, fixed! :)

api/src/run.ts
```diff
@@ -0,0 +1,67 @@
...
+import { opAsync, opSync } from "./utils.ts";
+
+// A generic job that we receive from Rust
+type AcceptedJob =
+    | { type: "http"; request: HttpRequest; responseRid: number }
+    | { type: "kafka"; event: KafkaEvent };
+
+// This is the entry point into the TypeScript runtime, called from `main.js`
+// with structures that describe the user-defined behavior (such as how to
+// handle HTTP requests)
```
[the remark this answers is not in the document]
honzasp: I added a comment.

api/src/routing.ts
honzasp: It was this feature:
https://www.notion.so/chiselstrike/Routing-afbeb5a48dda47a28f919f5a7165f387#75745f4d44ff4fb597f4f394d8414c6e

But I decided to remove it, it does not make any sense.

7flash (approved):
__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :trophy: This code is awesome

7flash (dismissed):

server/src/datastore/meta/mod.rs
```diff
@@ -421,76 +409,84 @@ impl MetaService {
...
         Ok(())
     }
 
-    /// Load the existing endpoints from from metadata store.
-    pub async fn load_sources<'r>(&self) -> anyhow::Result<PrefixMap<String>> {
-        let query = sqlx::query("SELECT path, code FROM sources");
+    /// Load module source codes from metadata store.
+    pub async fn load_modules(&self, version_id: &str) -> Result<HashMap<String, String>> {
+        let query =
+            sqlx::query("SELECT url, code FROM modules WHERE version = $1").bind(version_id);
```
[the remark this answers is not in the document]
honzasp: Yes, the `modules` table will be automatically created by the code in `server/src/datastore/meta/schema.rs`.

server/src/datastore/dbconn.rs
```diff
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: © 2021 ChiselStrike <info@chiselstrike.com>
+// SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
```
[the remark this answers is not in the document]
honzasp: It's just me being too eager, I will revert these changes.

honzasp:
I reverted the new CRUD from this branch, the changes now live on another branch:
https://github.com/honzasp/chiselstrike/pull/1

server/src/worker.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    /// Channel for signaling that the worker is ready to handle jobs.
+    ///
+    /// Once the worker sends the signal, this is reset to `None`.
+    pub ready_tx: Option<oneshot::Sender<()>>,
+
+    /// Channel for receiving jobs.
+    ///
+    /// We can get away with `Rc<RefCell<>>`, because the worker runs on a single thread.
+    pub job_rx: Rc<RefCell<mpsc::Receiver<VersionJob>>>,
```
dekimir, line 69: Why an `Rc`?  Doesn't each WorkerState get its dedicated receiver, separate from all others?  That's how I interpret the worker-spawning loop in `version::run()`.

dekimir (changes requested):

api/src/http.ts
```diff
@@ -0,0 +1,179 @@
...
+            }`,
+        );
+    }
+
+    // the HTTP request usually specifies only path and query, but we need a full URL; so we resolve the URL
+    // from the request with respect to an arbitrary base
+    const url = new URL(httpRequest.uri, location.href);
+
+    // initialize the legacy global request context
+    // note that this means that we can only handle a single request at a time!
```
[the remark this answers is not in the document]
honzasp: This problem runs much deeper than this function, it is inherent to the current design of our data layer (or, rather, lack of design).

server/src/ops/job.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[deno_core::op]
+async fn op_chisel_accept_job(
+    state: Rc<RefCell<deno_core::OpState>>,
+) -> Result<Option<AcceptedJob>> {
+    let job_rx: Rc<RefCell<_>> = state.borrow().borrow::<WorkerState>().job_rx.clone();
+
+    #[allow(clippy::await_holding_refcell_ref)]
+    async fn recv_job(job_rx: &RefCell<mpsc::Receiver<VersionJob>>) -> Result<Option<VersionJob>> {
+        // yes, we really *do* want to hold a `RefMut` across `.await` ...
```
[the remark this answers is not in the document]
honzasp: `RefMut`

server/src/ops/job.rs
```diff
@@ -0,0 +1,94 @@
...
+        match job_rx.try_borrow_mut() {
+            Ok(mut job_rx) => Ok(job_rx.recv().await),
+            // ... if somebody else tries to borrow this `RefCell` while we are `.await`-ing, they will
```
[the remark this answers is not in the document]
honzasp: Correct

server/src/worker.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub transaction: Option<TransactionStatic>,
+
+    /// Channel for signaling that the worker is ready to handle jobs.
+    ///
+    /// Once the worker sends the signal, this is reset to `None`.
+    pub ready_tx: Option<oneshot::Sender<()>>,
+
+    /// Channel for receiving jobs.
+    ///
+    /// We can get away with `Rc<RefCell<>>`, because the worker runs on a single thread.
```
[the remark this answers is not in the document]
honzasp: Because the `Arc<Version>`-s _are_ shared between threads:
1. The `Arc<Version>` is accessed from multiple workers
2. The `Arc<Version>` is accessed from HTTP and other "server-wide" code (that's why it needs to be stored in the `Trunk`)

server/src/version.rs
```diff
@@ -0,0 +1,143 @@
...
+    mut job_rx: mpsc::Receiver<VersionJob>,
+) -> Result<()> {
+    let ready_rxs = FuturesUnordered::new();
+    let mut job_txs = Vec::new();
+    let worker_handles = FuturesUnordered::new();
+
+    // spawn all workers for this version
+    for worker_idx in 0..init.worker_count {
+        let (ready_tx, ready_rx) = oneshot::channel();
+        let (job_tx, job_rx) = mpsc::channel(1);
```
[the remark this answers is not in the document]
honzasp: This _shadows_ the variable declaration from line 79 (not line 65, which is in a different function). The reason is simply that I was lazy and overloaded the name `job_rx` for the channel that receives jobs for the version and for one particular worker. I will rename the variable to make this more clear.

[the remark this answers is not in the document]
honzasp: The same applies to `ready_{tx,rx}`.

server/src/worker.rs
honzasp: The `Rc` is necessary for the `op_chisel_accept_job()`: I need to hold the `RefMut<mpsc::Receiver>` that I borrowed from `RefCell<mpsc::Receiver>` across await, so I need to hold a reference to the `RefCell` itself across await.

If the `RefCell<mpsc::Receiver>` was stored directly in the `WorkerState`, that would mean that I would have to hold the `Ref<WorkerState>` borrowed from the `deno_core::OpState` across the await too, so any other `op` that tried to access the `WorkerState` would fail.

---

In more detail:

Without the `Rc`, the "borrow chain" in `op_chisel_accept_job` would look like this:
- `&mut mpsc::Receiver` is mut-borrowed in `job_rx.recv().await` from
- `RefMut<mpsc::Receiver>` which is `RefCell.borrow_mut()`-borrowed in `match job_rx.try_borrow_mut() {` from
- `&RefCell<mpsc::Receiver>` which is passed to `recv_job()` and borrowed from
- `&WorkerState` which is borrowed from
- `Ref<WorkerState>` which is `RefCell.borrow()`-borrowed from
- `&deno_core::OpState` which is `RefCell.borrow()`-borrowed from
- `Rc<RefCell<deno_core::OpState>>` which is passed into `op_chisel_accept_job()`

So we would need to hold multiple `Ref`-s across the await, and any other op that would try to borrow the `OpState` or the `WorkerState` would fail.

But with `Rc<RefCell<mpsc::Receiver>>`, the "borrow chain" looks like this:
- `&mut mpsc::Receiver` is mut-borrowed in `job_rx.recv().await` from
- `RefMut<mpsc::Receiver>` which is `RefCell.borrow_mut()`-borrowed in `match job_rx.try_borrow_mut() {` from
- `&RefCell<mpsc::Receiver>` which is passed to `recv_job()` and borrowed from
- `Rc<RefCell<_>>` which we have _cloned_ into a local variable `job_rx` in `op_chisel_accept_job()`.

Thus, the `Rc` allows us to end the borrow chain before it reaches the `OpState`, so we don't have to keep any borrows from `OpState` alive across the await.

api/src/http.ts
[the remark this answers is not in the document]
dekimir: But this function's blurb is an ideal place to warn all readers of this problem.  Anyone thinking about handling requests will naturally look to `handleHttpRequest()`.  Seeing a warning comment right up top lowers the risk of a nasty bug in the future.

server/src/worker.rs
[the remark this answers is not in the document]
dekimir: Ok.

server/src/version.rs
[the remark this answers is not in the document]
dekimir: Ah, OK.  Sorry, I guess I was blind to the `let` part. :smiley:

server/src/worker.rs
dekimir: OK, but why must `WorkerState` hold on to its copy?  Can we just move the `Receiver` out of `WorkerState` while we need it, then move it back in when we're finished?

server/src/version.rs
```diff
@@ -0,0 +1,143 @@
...
+        Ok(())
+    }));
+
+    let version_id = version.version_id.clone();
+    let job_task = TaskHandle(task::spawn(async move {
+        // distribute jobs among workers in a round-robin fashion
+        // TODO: we should perhaps be more clever than round-robin
+        let mut worker_i = 0;
+        while let Some(job) = job_rx.recv().await {
+            if job_txs[worker_i].send(job).await.is_err() {
```
dekimir, line 124: Come to think of it, how does this code know that `worker_i` has finished the previous request?

api/src/http.ts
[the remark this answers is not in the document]
honzasp: Well, `handleKafkaEvent` has the same problem; perhaps I should add the comment to `run.ts`?

server/src/worker.rs
honzasp: Hohoho, that's a good idea! :) This would elegantly get rid of all the tricky code, and the overhead of moving the `mpsc::Receiver` back and forth is negligible in this case. I'll look at that in more detail and change the code, if there is not some problem that I cannot see right now.

server/src/version.rs
honzasp: It does not, if the `job_txs[worker_i]` channel is already full, this call will block (there is an `.await`). Of course, this naive round-robin is rather dumb and could certainly be improved, but I went for the simplest solution now and left the optimizations as future work, to avoid introducing extra complexity to this PR :)

dekimir: Good! :)  But that makes it essential for the channel to be of size 1.  I'll leave a comment in version.rs to ask for an explanation there.

server/src/version.rs
dekimir, line 88: Please add a comment here saying that size must be 1 or rivers of blood will flow (otherwise a worker will attempt to interleave handling of multiple requests and mess up horribly and unpredictably).

api/src/http.ts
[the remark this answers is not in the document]
dekimir: I'd start with big warnings atop `handleHttpRequest` and `handleKafkaEvent`.  Then you can additionally comment in run.ts saying "please never send me the next request before I complete the current one".

server/src/version.rs
honzasp: Why would the size 1 be essential? It would work the same if the size was 2 or 10, the size of the channel affects only the number of jobs that can get buffered before `send()` blocks.

server/src/version.rs
honzasp: This is not true, the channel works as a queue: what is important is that the code in `run.ts` only handles a single job at a time, which is trivially true, because there is a simple `loop { accept job; handle job }` with no concurrency. The number of jobs potentially waiting in the channel to be accepted and handled is irrelevant. (In fact, I would prefer a "rendez-vous" channel of size 0, but Tokio unfortunately does not support that.)

dekimir: > the code in `run.ts` only handles a single job at a time, which is trivially true, because there is a simple `loop { accept job; handle job }` with no concurrency.

I don't believe that's true.  Doesn't every `await` in that code (and there are three) create concurrency?  The runtime doesn't freeze and do nothing while waiting for a promise to resolve.

dekimir: Actually, I take that back.  @MarinPostma reminds me that `await` is precisely what prevents concurrency here.  So you're right, the loop in `run.ts` uses carefully placed `await`s to ensure there's no interleaving even if the channel size is larger than 1.  And if someone accidentally forgets an `await` there in a future update, the compilation will likely fail.

OK, there's no need for a comment on channel size here.

api/src/run.ts
```diff
@@ -0,0 +1,69 @@
...
+    specialAfter(routeMap);
+    const router = new Router(routeMap);
+
+    const topicMap = userTopicMap ?? new TopicMap();
+
+    // signal to Rust that we are ready to handle jobs
+    Deno.core.opSync("op_chisel_ready");
+
+    for (;;) {
+        const job = await opAsync(
```
dekimir, line 40: OK, let's definitely add a comment here that interleaved calls to `op_chisel_accept_job` aren't allowed until the job completes, so never forget to `await` everywhere.

server/src/ops/job.rs
```diff
@@ -0,0 +1,94 @@
...
+/// This is passed to JavaScript along with the request (in `AcceptedJob::Http`), and JavaScript
+/// then passes the response back to us by calling `op_chisel_http_respond`.
+struct HttpResponseTxResource {
+    response_tx: RefCell<Option<oneshot::Sender<HttpResponse>>>,
+}
+
+impl deno_core::Resource for HttpResponseTxResource {}
+
+#[deno_core::op]
+async fn op_chisel_accept_job(
```
dekimir, line 38: Here, too: let's add a blurb that cautions against calling this function repeatedly without waiting for the job to complete.

honzasp: Ok, that's probably a good idea; but note that if somebody mistakenly does call the function repeatedly, nothing catastrophic will happen, they will simply get a (very descriptive) runtime error.

api/src/run.ts
honzasp: Well, this is not technically true: `op_chisel_accept_job` only requires that you won't call it multiple times before it returns, it does not care whether the job completed or not. If you try to start two concurrent jobs, you will get an error from `op_chilsel_begin_transaction`, because each job will try to begin the global transaction.

honzasp: (To verify that `op_chisel_accept_job` is not called concurrently, the only conditions that need to be true are:
1. `run()` is only called once
2. `op_chisel_accept_job` is only called from `run()` and the call is immediately `await`-ed)

honzasp:
I think that I have addressed all issues brought up in review.

dekimir (changes requested):
Just a couple more, before I forget...

server/src/trunk.rs
```diff
@@ -0,0 +1,112 @@
...
+struct TrunkState {
+    versions: HashMap<String, TrunkVersion>,
+    tasks: FuturesUnordered<CancellableTaskHandle<Result<()>>>,
+    waker: Option<Waker>,
+}
+
+#[derive(Clone)]
+pub struct TrunkVersion {
+    pub version: Arc<Version>,
+    /// NOTE: this sender cannot be stored in the `Version`, because the version terminates only
```
dekimir, line 32: But doesn't Version happily hold all Workers' senders without facing this problem?  What's the difference?

And why doesn't Trunk spawn Versions, like Version spawns Workers?  Instead we have `server` and `rpc` spawning versions and immediately calling `trunk.add_version` on the result.

server/src/trunk.rs
```diff
@@ -0,0 +1,112 @@
...
+        // if there is still a task in `state.tasks` for this version, we just leave it alone. it
+        // should terminate on its own when all `Arc<Version>`s are dropped
+        state
+            .versions
+            .remove(version_id)
+            .map(|trunk_version| trunk_version.version)
+    }
+}
+
+pub async fn spawn() -> Result<(Trunk, TaskHandle<Result<()>>)> {
```
dekimir, line 93: Can there be a Trunk without a task?  Can a Trunk have multiple tasks for it?  If not, I don't see why we separate the task from the Trunk.

And same goes for Version and Worker.

server/src/trunk.rs
```diff
@@ -0,0 +1,112 @@
...
+        // documentation of `FuturesUnordered` for details)
+        if let Some(waker) = state.waker.take() {
+            waker.wake()
+        }
+    }
+
+    pub fn remove_version(&self, version_id: &str) -> Option<Arc<Version>> {
+        let mut state = self.state.write();
+        // if there is still a task in `state.tasks` for this version, we just leave it alone. it
+        // should terminate on its own when all `Arc<Version>`s are dropped
```
dekimir, line 85: This seems to contradict line 34.

server/src/trunk.rs
honzasp: No, `Version` does not reference any sender, directly or indirectly.

The `job_task` task that runs as part of the "version task tree" owns the version job receiver and the worker job senders. When the version job sender is dropped, the `job_task` will terminate and the worker job senders will be dropped, so the worker job receivers will become closed and the workers will terminate.

During apply, we need to spawn a "canary" version to check whether the provided code is correct; this version must not be added to the `Trunk`, because it is not supposed to replace the current version.

However, as you mention elsewhere, we spawn versions and add them to trunk in at least three different places: it might be good idea to create a function that does this and make it a part of the `Trunk` code. I will investigate this after I'm back from the vacation.

server/src/trunk.rs
honzasp: This pattern is used in multiple places in the code (server, trunk, version, worker): there is a task that performs some work, and an object that contains state related to this work.

We cannot store the `TaskHandle` inside of the `Trunk`, because we need to store the `Trunk` in the `Server` structure, but we need to `.await` (consume) the task in `server::run()`.

server/src/trunk.rs
honzasp: Good catch, I didn't update the comment when I updated the code.

server/src/server.rs
```diff
@@ -1,513 +1,317 @@
...
+    // Note that we spawn many tasks, but we .await them all at the end; we never leave a task
+    // running in the background. This ensures that we handle all errors and panics and also that
+    // we abort the tasks when they are no longer needed (e.g. if other task has failed).
+    //
+    // This approach is called "structured concurrency", and it seems to be a good way to write
+    // concurrent programs and keep your sanity.
+
+    let (server, trunk_task) = make_server(opt).await?;
+    start_versions(server.clone()).await?;
+    start_chiselstrike_version(server.clone()).await?;
```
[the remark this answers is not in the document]
honzasp: We don't have an explicit concept of "user" vs "internal" version; in fact, it seems to me that the `__chiselstrike` version is really a kludge, which is the reason why the code looks kludge-y. (Unclear code and unclear prose are both the result of unclear thinking.)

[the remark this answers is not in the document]
dekimir: We definitely have types and endpoints that the user never defined, so they go into a special version.  Maybe you don't want to call that "an explicit concept", but we do have that.  And that special version's name is arbitrary -- any name is OK, as long as it doesn't clash with the user's versions.  I question the decision to use that name here in this function, because the result is confusing -- are versions from line 53 somehow not ChiselStrike versions??  I agree about unclear code/prose (why must all types -- even builtins -- belong to a version?), but this naming decision reduces clarity further.  Can't we name these functions something better?

server/src/trunk.rs
dekimir: > No, `Version` does not reference any sender, directly or indirectly.

Ah, I see: `version::run()` does, but not the `Version` struct.  I guess the difference is that `Trunk` has a simpler job than `Version`: it just needs to look up a version id and give the corresponding sender to the HTTP request handler.  `Version`, in contrast, has to distribute jobs to workers.

> During apply, we need to spawn a "canary" version to check whether the provided code is correct; this version must not be added to the `Trunk`, because it is not supposed to replace the current version.

I wonder if we truly need to _spawn_ a version just to check the code.  It doesn't sound right to create tasks you don't ever intend to activate.  Perhaps there should be common code at a finer granularity between code-check and actual spawning?

server/src/trunk.rs
dekimir: Is that to say that `Trunk` must outlive its task?

It's good that the same pattern is used everywhere, it makes it easier to read the code; but the pattern is open to misinterpretation -- the task and the state object are separate, even though they're clearly related in non-trivial ways.  We should describe that relation explicitly, either through design (preferably) or through comments.

Also, if these structs are state, why aren't they all named with a "State" suffix, like WorkerState is?  It feels like Server, Trunk, and Version are not a task's state but legit APIs on their own, with clients that don't care about the ongoing tasks.

server/src/trunk.rs
honzasp: The canary version is the same as any other version, it is just not added into the `Trunk`. To check whether the code is correct, we need to spawn the Deno tasks and everything, so we cannot really simplify the spawning of canary versions too much.

I'm not sure what you mean by "activating" the tasks?

server/src/trunk.rs
honzasp: You are right, I'm not very consistent here, `Server` could as easily be named `ServerState`. I think that the confusion comes from the fact that we are used to object-oriented programming and objects that _do_ things. But objects are just data structures, the tasks are "doing things".

`Server` stores server-wide data structures and the server tasks perform the work of the server (such as receiving HTTP requests or running Deno), `Trunk` is a data structure that stores all active versions and their tasks, and the trunk task polls the version tasks, etc.

I'm not sure how we can describe the relation between the `Trunk` and the trunk task more explicitly than it is now, when both pieces of data are returned by `trunk::spawn`. Could you describe what exactly you find unclear, what confusion might arise?

server/src/server.rs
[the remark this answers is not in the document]
honzasp: The versions from line 53 are ChiselStrike versions, but they are not `__chiselstrike` versions.

The concept of an "internal version" and "types and endpoints" that the user never defined are quite weird. Versions are completely independent in code and in data, _except_ that they use the same database for storing users and sessions. This "design decision" means that, for example, we cannot use SQLite for development versions and Postgres for production versions, because all versions must be in the same database!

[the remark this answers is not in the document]
dekimir: > The versions from line 53 are ChiselStrike versions, but they are not `__chiselstrike` versions.

... and the name `__chiselstrike` is arbitrary and inconsequential.  Please don't use it here like this, it's harmful to the reader's understanding and entirely unnecessary.  Use anything else: `internal`, `auth`, `universal`, `special`, `intrinsic`, `builtin` -- anything but `chiselstrike`.

> "types and endpoints" that the user never defined are quite weird. Versions are completely independent in code and in data, _except_ that they use the same database for storing users and sessions.

Agreed, and we can change that later.  But until we change it, let's please not confuse the reader any further.

server/src/trunk.rs
dekimir: > I'm not sure what you mean by "activating" the tasks?

I mean tasks never receiving any jobs and getting dropped as soon as they signal they're ready.

dekimir: > To check whether the code is correct, we need to spawn the Deno tasks and everything

Tell me more about this.  Why isn't just compiling the module code enough to validate it?  Do we need to execute it to build a RouteMap?

server/src/trunk.rs
dekimir: > I'm not sure how we can describe the relation between the `Trunk` and the trunk task more explicitly than it is now, when both pieces of data are returned by `trunk::spawn`. Could you describe what exactly you find unclear, what confusion might arise?

Is a Trunk object allowed to have zero tasks?  How about multiple tasks?  Does a Trunk ever outlive its task or vice versa?  If I have a Trunk object and want to get to its task(s), how do I do that?  How do I do the opposite?

dekimir: Actually, I'm noticing that `TrunkState` is a separate struct.  That's... complicated.  I don't think I understand the pattern you're talking about.

server/src/trunk.rs
honzasp: Yes, just importing the modules would only check for syntax errors, initializing the modules and building the `RouteMap` catches muchmore errors.

server/src/trunk.rs
honzasp: In the case of `Trunk`, I had to introduce `TrunkState` because I need to wrap it in the `Arc<RwLock>`. I agree that the naming is not very consistent, do you have some suggestion for better naming?

> Is a Trunk object allowed to have zero tasks? How about multiple tasks?
> If I have a Trunk object and want to get to its task(s)

When you "spawn a trunk", you get a `Trunk` instance and a handle to the task. To be honest, it never even occured to me that the relationship would be unclear, and I still don't understand what you don't understand :)

> Can I move or drop a Trunk object while its task is active?

As you can see, the `Trunk` object is just a wrapper around `Arc<Mutex>`, so you can clone it and drop it as you wish, without affecting anything else.

> How do I deduce the task's state from a Trunk instance?

I don't understand this question. What does "deducing a task's state" mean and why would you want it?

> why does Trunk have methods?

The methods of the `Trunk` have methods to manipulate the `Trunk` data structure.

> Version, in fact, has two info objects: Version and VersionInfo

`VersionInfo` is a legacy struct whose meaning eludes me, but I didn't want to remove it to avoid being accused of making the PR too big.

honzasp: I mean, you really don't understand how the code works, or you understand it but think that it is badly designed? If the former is true, maybe I should explain it more? If the latter, then I agree that the naming is inconsistent and I would welcome any suggestions for improvement.

server/src/trunk.rs
honzasp: > I mean tasks never receiving any jobs and getting dropped as soon as they signal they're ready.

I see; why doesn't this sound right to you? Do you think that introducing extra code would be better?

server/src/server.rs
```diff
@@ -1,513 +1,317 @@
...
+        export * from 'chisel:///chiselstrike_root.ts';
+        "
+        .into(),
+    );
 
-    let start_wait = async move {
-        for _id in 0..opt.executor_threads {
-            readiness_rx.recv().await.unwrap();
-        }
+    let (ready_tx, _ready_rx) = oneshot::channel();
```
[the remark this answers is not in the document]
honzasp: Unfortunately, I was not able to find a meaningful drying.

server/src/trunk.rs
dekimir: After chatting in person, we agreed it's fine to validate modules by spawning a trial-balloon version and shutting it down as soon as it's ready to receive tasks.  This ensures the real spawning will also succeed, and it doesn't prevent future evolution of real spawning.

server/src/trunk.rs
dekimir: After chatting in person, we agreed this is a comment/doc issue and shouldn't hold up this PR.  We'll address it separately later.

dekimir (commented):
LGTM.  (Not formally approving yet to prevent automerge.)

penberg (approved):

PR #1621, opened 2022-08-05
Inspector improvements

### LINKED ISSUE #1086: Make it possible to enable inspector for `chisel dev` and document it
Alex Ausch writes:

> also, is there a way for me to inject breakpoints and step through / debug?

Let's make it possible to enable the inspector for `chisel dev` and document it too.

### LINKED ISSUE #1620: Cannot integrate ChiselStrike project as Chrome Inspector workspace
If I add ChiselStrike project to Chrome Inspector as workspace, at least breakpoints have no effect. I suspect we need to do some more work in providing enough metadata for this to work.

penberg:
This pull request improves the inspector to make it more usable to
developers:

* Change `--inspect-brk` semantics to match Deno.
* Add an `--inspect` flag that does not wait for a debugger.
* Make it easy to enable inspector with `chisel dev` and document it.

Fixes #1086.

Please note that we need to fix #1620 as future work to fix the
remaining issues with the inspector support.


cli/src/main.rs
```diff
@@ -65,6 +65,9 @@ enum Command {
         /// calls tsc --noEmit to check types. Useful if your IDE isn't doing it.
         #[structopt(long)]
         type_check: bool,
+        /// Activate inspector and let a debugger attach at any time.
+        #[structopt(long)]
+        inspect: bool,
```
MarinPostma, line 70: this seems a bit redundant, since we can now pass args to chiseld from the cli now (anything after `--` is passed to chiseld)

penberg: @MarinPostma It's redundant, but the commit message explains why we want it (`npm run dev` becomes prettier).

MarinPostma (approved):

MarinPostma (approved):

PR #1625, opened 2022-08-08
Introduce rust-based framework for integration tests

BearLemma:
PR ready to be merged. 
It introduces new class of integration tests that are based on our custom Rust framework.


cli/tests/integration_tests/rust_tests/test_bad_filter.rs
```diff
@@ -0,0 +1,34 @@
+use crate::framework::{IntegrationTest, OpMode, TestConfig};
+
+inventory::submit!(IntegrationTest {
+    name: "test_bad_filter",
+    mode: OpMode::Deno,
+    test_fn: &test_bad_filter
+});
```
honzasp, line 7: I hate this :D Why not register the tests in `mod.rs`? A simple case could look like this:

```
suite.register_deno("test_bad_filter", test_bad_filter::test);
```

but it could be easily extended, for example to run the test only with optimizations:

```
suite.register_deno("test_bad_filter", test_bad_filter::test).optimize(true);
```

or to run the same test with multiple parameterizations:

```
suite.register_deno("test_bad_filter_a", |cfg| test_bad_filter::test(cfg, "a"));
suite.register_deno("test_bad_filter_b", |cfg| test_bad_filter::test(cfg, "b"));
```

cli/tests/integration_tests/rust_tests/test_bad_filter.rs
```diff
@@ -0,0 +1,34 @@
...
+
+pub async fn test_bad_filter(config: TestConfig) {
+    let mut ctx = config.setup_env().await;
+    let (chisel, _chiseld) = ctx.get_chisels();
+
+    chisel.copy("examples/person.ts", "models");
```
honzasp, line 13: Do we want to depend on the working directory for this? Maybe we should use `include_str!` instead and not read from the filesystem at all?

cli/tests/integration_tests/rust_tests/test_find_by.rs
```diff
@@ -0,0 +1,147 @@
...
+            "/dev/find_by",
+            json!({
+                "field_name":"height",
+            }),
+        )
+        .await;
+    assert_eq!(
+        resp_txt,
+        "Glauber Costa 666 true 10.01 Jan Plhak -666 true 10.02 "
+    );
```
honzasp, line 130: Can't we make this a little less verbose? In particular, without the uglifier, the code could be much more compact:

```
let resp_txt = chisel.post_text("/dev/find_by", json!({
    "field_name":"height",
})).await;
assert_eq!(resp_txt,
    "Glauber Costa 666 true 10.01 Jan Plhak -666 true 10.02 ");
```

honzasp (commented):
I left some preliminary remarks and suggestions :) But these are details, I think that the overall direction is good!

cli/tests/integration_tests/rust_tests/test_find_by.rs
MarinPostma: you can add `#[rustfmt_skip] ` if that's what you want

honzasp (approved):

PR #1630, opened 2022-08-10
add test proc macro

MarinPostma:
This pr adds macros to declare rust integration tests in the cli crate.


chisel-macros/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "chisel-macros"
```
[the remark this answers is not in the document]
MarinPostma: https://github.com/rust-lang/cargo/issues/5042

I think not unfortunately.

But this can be the home for other macros in the future

cli/tests/integration_tests/chisel-macros/src/lib.rs
```diff
@@ -0,0 +1,84 @@
+use std::path::PathBuf;
+
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input, Ident, ItemFn, LitStr, Path, Token,
+};
+
+struct TestArgs {
```
dekimir, line 10: Can we please have a short blurb here hinting at what this is and how to use it?  And same for the fields below.

dekimir (commented):
Looks good, just one nitpick.

cli/tests/integration_tests/chisel-macros/src/lib.rs
MarinPostma: you have it!

cli/tests/integration_tests/chisel-macros/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input, Ident, ItemFn, LitStr, Path, Token,
+};
+
+/// Arguments to the test macro
+struct TestArgs {
+    /// Mode in which to run the tests: either OpMode::Deno or OpMode::Node.
```
dekimir, line 12: But why is the type not `OpMode` then?

cli/tests/integration_tests/chisel-macros/src/lib.rs
```diff
@@ -0,0 +1,86 @@
+use std::path::PathBuf;
+
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input, Ident, ItemFn, LitStr, Path, Token,
+};
+
+/// Arguments to the test macro
```
dekimir, line 10: Heh, this doesn't tell me much.  (And, in fact, suggests that `TestMacroArgs` is a better struct name.)

I was hoping for something like "Describes how a test should be run" or "Allows control over test behavior when the test is executed".

cli/tests/integration_tests/chisel-macros/src/lib.rs
MarinPostma: I am in the context of a proc macro here, all I know about are tokens. I am matching against something that's a `syn::Path`, and then injecting it in the code here: https://github.com/chiselstrike/chiselstrike/pull/1630/files#diff-62854406ae6accfc6e72d221a2861d6773d3b06a85edc846a0f5cca486246d24R53.

cli/tests/integration_tests/chisel-macros/src/lib.rs
MarinPostma: This struct is only there to collect the arguments to the macro that I parse from a plain `TokenStream`, this is not exposed to the consumer of the macro. Am I right to think that what you actually want is the documentation for the macro itself?

cli/tests/integration_tests/chisel-macros/src/lib.rs
dekimir: So then this isn't supplied as an OpMode but rather something that parses into OpMode?  If so, that's not obvious from the blurb and code.

cli/tests/integration_tests/chisel-macros/src/lib.rs
dekimir: The macro having a blurb can't be a bad thing. :)  As for this struct, if it's just a collection of Paths, ... well, perhaps the macro should just take `Vec<Path>`?

MarinPostma: It's not a collection of `Path`. This struct contains all the arguments that can be passed to the `test` macro. Currently, the only argument that is passed to it is the `mode` argument:
```rust
use OpMode::Deno;
#[chisel_macros::test(mode = Deno)]
async fn my_test() { .. } 
```
or
```rust
#[chisel_macros::test(mode = OpMode::Deno)]
async fn my_test() { .. } 
```

this gets expanded to :
```rust
inventory::submit! {
            IntegrationTest {
                name: "my_test",
                mode: OpMode::Deno, // or whatever you passed to the macro.
                test_fn: &my_test,
            }
        }

async fn my_test() { .. } 
```

Now it happens that `OpMode::Deno` or `Deno` or `crate::framework::Opmode::Deno` parse into a [syn::Path](https://docs.rs/syn/latest/syn/struct.Path.html). If I'm adding a new argument tomorrow, there's no reason to assume that it'll parse to a `syn::Path`, it could be a boolean:

```rust
#[chisel_macro::test(mode = OpMode::Deno, optimized = true)]
async fn my_test() {...}
```

if which case my struct would become:
```
struct TestArgs {
    mode: Path,
    optimized: Lit,
}
```

PR #1631, opened 2022-08-10
granular readiness

glommer:
We always return 200 on the readiness probe, but there are some race
conditions here, due to which it is still possible for the readiness
internal health endpoint to say we are ready to take requests when
in reality we are not.

This increases the granularity and gives us more control over how
we declare whether or not we're ready.

### DIFF
--- a/server/src/internal.rs
+++ b/server/src/internal.rs
@@ -8,9 +8,19 @@ use once_cell::sync::OnceCell;
 use serde_derive::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::net::SocketAddr;
+use std::sync::atomic::{AtomicU16, Ordering};
 
 /// If set, serve the web UI using this address for gRPC calls.
 static SERVE_WEBUI: OnceCell<SocketAddr> = OnceCell::new();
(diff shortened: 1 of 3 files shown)


penberg (approved):

dekimir:
Hm, should `chisel wait` wait for `/readiness` instead of rpc_addr to start listening?  If not, then I'm unsure what the difference is.

dekimir:
One problem is that `chisel` options don't seem to include the API listen address:
https://github.com/chiselstrike/chiselstrike/blob/ec67428fd41e02640004e52f36e552d187a91cf8/cli/src/main.rs#L36-L42

We can add that option, but that will break existing uses of `chisel wait -r`.

penberg:
@dekimir Aah, right. I guess it's better to just fix the RPC because `chisel` clients might not even have access to the internal API because of firewalls and network configuration.

dekimir:
Does RPC even need fixing?  It might very well _not_ have the race condition that this PR fixes.  Isn't it true that when RPC listener is up, `chiseld` is up?

PR #1632, opened 2022-08-11
CRUD: read object before PUT

### LINKED ISSUE #1213: Add crud()'s support to PATCH requests
As discussed on this [issue](https://github.com/chiselstrike/chiselstrike/issues/1207), we have implemented PUT requests that fully replace an entity. Now we need PATCH requests that should applies partial modifications.

One thing to think about is: If I have a optional field set and want to change it to `undefined`, what should I pass to the request? Using `PUT` I simply don't pass the field at the body, but if I do this using `PATCH` the field doesn't change

glommer:
This is crucial for objects with non-literal defaults. Consider for
example, the following object:

```
export class Joke extends ChiselEntity {
        createdAt : number = Date.now()
        name : string
        content : string
}
```

because "createdAt" has a default, it is convenient not to specify it
when creating the object. However, if you call a CRUD PUT on an existing
object, the whole object gets recreated and have the default it had
before (in this case, a creation time).

With a PATCH, we can pass just name and content, and createdAt will be
left alone.

Fixes #1213

### DIFF
--- a/cli/tests/integration_tests/lit_tests/crud.node
+++ b/cli/tests/integration_tests/lit_tests/crud.node
@@ -218,16 +218,28 @@ $CURL -X DELETE "$CHISELD_HOST/dev/persons?all=true"
 # CHECK: HTTP/1.1 200 OK
 # CHECK: Deleted entities matching ?all=true
 
-$CURL -X PUT -d '{"age":98422}' $CHISELD_HOST/dev/persons/cef5d492-d7e3-4c45-9a55-5929b9ab8292
+$CURL -X PUT -d '{"age":98422, "height": 2 }' $CHISELD_HOST/dev/persons/cef5d492-d7e3-4c45-9a55-5929b9ab8292
 # CHECK: HTTP/1.1 200 OK
 # CHECK: "age": 98422
-$CURL -X PUT -d '{"age":2938}' $CHISELD_HOST/dev/persons/cef5d492-d7e3-4c45-9a55-5929b9ab8292 # Repeated PUT of the same id.
+# CHECK: "height": 2
+$CURL -X PUT -d '{"age":2938}' $CHISELD_HOST/dev/persons/cef5d492-d7e3-4c45-9a55-5929b9ab8292 # Repeated PUT of the same id, object is entirely recreated
 # CHECK: HTTP/1.1 200 OK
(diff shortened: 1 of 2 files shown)


dekimir:
I'd rather fix #1213 than pollute PUT.

glommer:
@dekimir I think you are right about this.

dekimir:
> @dekimir I think you are right about this.

Then I'm turning this into a draft, so it doesn't accidentally get approved and merged.

cli/tests/integration_tests/lit_tests/crud.node
```diff
@@ -218,16 +218,25 @@ $CURL -X DELETE "$CHISELD_HOST/dev/persons?all=true"
...
 # CHECK: there is just 1 entry
 $CURL --no-include $CHISELD_HOST/dev/persons | jq '.results | .[] | .age'
 # CHECK: 2938
+$CURL -X PATCH -d '{"height": 4 }' $CHISELD_HOST/dev/persons/cef5d492-d7e3-4c45-9a55-5929b9ab8292 # PATCH on an existing object, only mentioned properties are changed
+# CHECK: HTTP/1.1 200 OK
+# CHECK: "age": 2938
+# CHECK: "height": 4
+
+$CURL -X PATCH -d '{"height": 4 }' $CHISELD_HOST/dev/persons/0000 # PATCH of a non-existent object, 404
+# CHECK: HTTP/1.1 404 Not Found
```
dekimir, line 239: What about no ID?

dekimir (approved):
LGTM, one nitpick.

dekimir:
> I forced pushed this with an implementation of PATCH and a fixes line.

Thank you.  And I updated this PRs description from your commit message.

glommer:
I can add a test for no id too, but should be the same as PUT. But if I don't do it fast enough this will get merged on your approval =)

BTw, the description is helpful, but for 1-patch PRs that description is lost once the PR is merged and the commit message is what survives anyway

glommer:
@dekimir force-pushed a test with no-id. You'll have to reapprove (maybe wait for tests to succeed first)

glommer:
@dekimir ready to go

dekimir (approved):

PR #1634, opened 2022-08-11
Kafka consumer support

penberg:
This adds initial support for consuming Kafka topics.

The API looks as follows:

* You write an `events/<topic>.ts` file where `topic` is the name of the 
  Kafka topic to consume.
* You then pass `--kafka-connection` and `--kafka-topics <topic>` on the
  `chiseld` command line.
* Now whenever there is a record on a topic, your event handler is
  called.

In order to keep the pull request manageable, I am deferring the
following as future work:

* Automatically subscribe to topics when `chisel dev` is running.
* Support for multiple Kafka connections.
* Re-connect to Kafka broker on failure
* Add event handlers to `chisel describe`
* Introspection support
* Integration test with Kafka broker


api/src/worker.ts
```diff
@@ -93,6 +102,25 @@ async function importEndpointsImpl(endpoints: [Endpoint]) {
...
+
+        // Modules are never unloaded, so we need to create an unique
+        // path. This will not be a problem once we publish the entire app
+        // at once, since then we can create a new isolate for it.
+        const url = `file:///${apiVersion}/events${path}`;
+        const mod = await import(url);
+        const handler = mod.default;
+        if (typeof handler !== "function") {
+            throw new Error(
+                "expected type `v8::data::Function`, got `v8::data::Value`",
```
honzasp, line 119: Hmm, why the reference to `rusty_v8` types? Perhaps a message like `"the default export from an event handler file must be a function"` would be more helpful?

api/src/worker.ts
```diff
@@ -256,6 +291,49 @@ function callHandler(
...
+async function callEventHandlerImpl(
+    path: string,
+    apiVersion: string,
+    rawKey: ArrayBuffer,
+    rawValue: ArrayBuffer,
+) {
+    await Deno.core.opAsync("op_chisel_start_event_handler");
+
+    // FIXME: maybe defer creating the transaction until we need one, to avoid doing it for
+    // endpoints that don't do any data access. For now, because we always create it above,
```
honzasp, line 303: endpoints -> event handlers

api/src/worker.ts
```diff
@@ -256,6 +291,49 @@ function callHandler(
...
+    // it should be safe to unwrap.
+    await Deno.core.opAsync("op_chisel_create_transaction");
+
+    const fullPath = "/" + apiVersion + path;
+
+    const key = new Uint8Array(rawKey);
+    const value = new Uint8Array(rawValue);
```
honzasp, line 310: What is the reason you decided to pass `Uint8Array` instead of an `ArrayBuffer` to the user code? The user can easily convert an `ArrayBuffer` into `Uint8Array` if they want to, but not the opposite: in general, an `Uint8Array` can be an arbitrary slice of an `ArrayBuffer`/ This is similar to passing `&[u8]` instead of `Vec<u8>` in Rust.

api/src/worker.ts
```diff
@@ -272,18 +350,29 @@ onmessage = function (e) {
             initWorker(d.id);
             break;
         case "importEndpoints":
```
honzasp, line 352: This event now works with endpoints and event handlers, it should perhaps be removed?

cli/src/project.rs
```diff
@@ -72,6 +73,8 @@ pub(crate) struct Manifest {
     pub(crate) models: Vec<String>,
     /// Vector of directories to scan for endpoint definitions.
     pub(crate) endpoints: Vec<String>,
+    /// Vector of directories to scan for event handler definitions.
+    pub(crate) events: Option<Vec<String>>,
```
honzasp, line 77: Why use the name "events" for a directory that contains event handlers? By the same logic, the "endpoints" directory would be named "requests".

cli/tests/integration_tests/rust_tests/test_kafka.rs
```diff
@@ -0,0 +1,20 @@
...
+    let (chisel, _chiseld) = ctx.get_chisels();
+
+    chisel.write(
+        "events/test-topic.ts",
+        r##"
+        import { ChiselEvent } from "@chiselstrike/api";
+
+        export default async function (event: ChiselEvent) {
+            console.log(msg);
+        }
```
honzasp, line 15: I wonder how this could typecheck, when the parameter is named `event` but the printed variable is `msg`?

docusaurus/docs/Intro/streaming.md
```diff
@@ -0,0 +1,27 @@
...
+
+To consume Kafka topics, you first need to write an `events/<topic>.ts` file that is the topic event handler function.
+
+For example, to consume events from a `hello` topic, create a `events/hello.ts` file with the following contents:
+
+```typescript title="my-backend/events/hello.ts"
+import { ChiselEvent } from "@chiselstrike/api";
+
+export default async function (event: ChiselEvent) {
+    console.log("Received an event from topic: " + JSON.stringify(event));
```
honzasp, line 15: This might be very confusing, because the JSON serialization of `Uint8Array` looks like this:
```
> JSON.stringify(new Uint8Array([10, 20, 30, 40]))
'{"0":10,"1":20,"2":30,"3":40}'
```

api/src/worker.ts
```diff
@@ -256,6 +291,49 @@ function callHandler(
...
+    // endpoints that don't do any data access. For now, because we always create it above,
+    // it should be safe to unwrap.
+    await Deno.core.opAsync("op_chisel_create_transaction");
+
+    const fullPath = "/" + apiVersion + path;
+
+    const key = new Uint8Array(rawKey);
+    const value = new Uint8Array(rawValue);
+
+    eventHandlers[fullPath]({ key, value });
```
honzasp, line 312: We pass raw bytes to the user code, so the user will have to deserialize the message. Perhaps we should provide some support code that will make this easier? I don't know what people send in their Kafka messages, though.

server/src/rpc.rs
```diff
@@ -252,8 +253,13 @@ impl RpcService {
                 let path = format!("/{}/{}", api_version, path);
                 endpoint_paths.push(path);
             }
+            if let Some(path) = path.strip_prefix("events/") {
```
honzasp, line 256: The event handlers don't have to be in directory named "events/", because the user can change the `events` field in their "Chisel.toml" file to point to any directory. (I think we should drop "Chisel.toml" and always use hardcoded paths.)

server/src/server.rs
```diff
@@ -156,14 +163,30 @@ pub(crate) async fn add_endpoints(
...
+        if path.contains("/endpoints/") {
+            let path = deno::endpoint_path_from_source_path(path);
+            activate_endpoint(&path).await?;
+
+            let func = Arc::new({
+                let path = path.to_string();
+                move |req| deno::run_js(path.clone(), req).boxed_local()
+            });
+            api_service.add_route(path.into(), func);
+        } else if path.contains("/events/") {
```
honzasp, line 176: Same as above, we cannot assume that endpoints and event handlers are stored in directories named "endpoints/" and "events/", because the user can change that in their "Chisel.toml".

honzasp (commented):
This looks reasonable, I left some comments in the code. I didn't look too deeply into the JavaScript interop madness in "deno.rs", "worker.ts" and "endpoint.ts", as I hope that #1618 will eventually land :)

There are two more general comments:

1. We should provide a way to register event handlers programmaticaly and not force the user to create a file for every topic. I guess that people often use patterns in topic names (see for example this blog post from some random guy [0]), so it might be useful to express this regularity in code.
2. I don't understand how Kafka ensures that messages are delivered at least once (or at most once) and how do we interact with this mechanism? For example, if the event handler throws an exception, will the message be delivered again?

[0] https://devshawn.com/blog/apache-kafka-topic-naming-conventions/#naming-structure

api/src/endpoint.ts
```diff
@@ -98,10 +98,17 @@ export async function readWorkerChannel() {
 
 type Endpoint = { path: string; apiVersion: string; version: number };
 
-export async function importEndpoints(endpoints: [Endpoint]) {
+type EventHandler = { path: string; apiVersion: string; version: number };
+
+// FIXME: rename to importModules()
```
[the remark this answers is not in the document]
penberg: @honzasp Please note that I added that comment before #1618 surfaced. I don't think renames make much sense as most of this code is hopefully going away.

penberg:
@honzasp The idea here is that the `events/<topic>.ts` convention is the default, simple way of consuming events. Fully agreed that we need more fine-grained API, but I want to defer that until we have the new routing API in place, so we can attempt to make them as unified as possible.

penberg:
@honzasp Kafka guarantees "at-most-once" semantics based on the offset you give it. In our case, we set the offset to be `StartOffset::Latest`, which means that Kafka will deliver the message to use once. If we fail to process it, we will never see it again. That's obviously not for all use cases, but it's not clear yet how to pass the offset. I was thinking we might need a `chisel events subscribe` command or something like that to decouple the subscription from server startup.

cli/src/project.rs
penberg: @honzasp Please propose an alternative name if you think there's a better one.

api/src/worker.ts
penberg: @honzasp This is copy-paste from `importEndpointsImpl()`. I don't want to change it unless we change both. And I don't want to change both because this code is going away.

server/src/api.rs
```diff
@@ -155,6 +159,7 @@ pub(crate) struct ApiService {
     // with runtime checking, which is likely cheaper, but still this is safer and we don't
     // have to manually implement Send (which is unsafe).
     paths: Mutex<PrefixMap<RouteFn>>,
+    event_handlers: Mutex<PrefixMap<EventFn>>,
```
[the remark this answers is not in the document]
penberg: Yup, I think `ApiService` needs a rename, but I don't know to what. It describes the external interfaces of the application that's running on the `chiseld` server.

docusaurus/docs/Intro/streaming.md
```diff
@@ -0,0 +1,27 @@
...
+}
+```
+
+Then, pass the `--kafka-connection <host>` and `--kafka-topics <topic>` command line options to the ChiselStrike server.
+
+For example, if your Kafka broker is running on `localhost:9092` and you want to subscribe to the `hello` topic, run:
+
+```
+npm run dev -- -- --kafka-connection localhost:9092 --kafka-topics test-topic
```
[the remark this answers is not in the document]
penberg: @honzasp ChiselStrike actually _does_ consume all topics automatically. If you did not define an event handler, a warning will be printed to the logs. I am actually not happy with the `--kafka-topics` command line options, but did not yet think of a better mechanism, so I would like to defer the issue of multiple topics for now.

api/src/worker.ts
penberg: @honzasp This is copy-paste from `callHandlerImpl()`. Dropped the comment altogether.

docusaurus/docs/Intro/streaming.md
penberg: @honzasp Fixed by tweaking the example.

server/src/rpc.rs
penberg: @honzasp Right. I looked into fixing this, but it's a pre-existing bug that looks non-trivial because the server has no knowledge of `Chisel.toml`. IOW, I think we should defer fixing this after the merge.

I opened a bug on this, thoough: https://github.com/chiselstrike/chiselstrike/issues/1650

server/src/server.rs
penberg: Makes sense, but I think should be fixed separately. Ticket here: https://github.com/chiselstrike/chiselstrike/issues/1650

cli/tests/integration_tests/rust_tests/test_kafka.rs
penberg: @honzasp You need to pass `--type-check` to `chisel apply` in Node mode if you want typechecking. Fixed the test case, though.

api/src/worker.ts
penberg: @honzasp There are various different formats that people use. We definitely want to make this part easier, but it's not yet clear what a good default is, which is why it's just blob for now.

cli/src/project.rs
honzasp: You already use the name "handler" in the code, so we should perhaps call the directory `handlers/`. On the other hand, Kafka uses the name "consumer" for this concept, but if I understand correctly, we want to make this mechanism more generic?

honzasp: Well, on second thoughts, `handlers/` is a bad name, because we have "event handlers" and "request handlers". Perhaps we should name it like this:
- If we keep the `endpoints/` directory for HTTP routes, keep the `events/` name (or `consumers/`?)
- If we use the `routes/` directory for HTTP routes, use `topics/`: `routes/user.ts` defines HTTP route `user`, and `topics/user.ts` defines Kafka topic `user`.

honzasp: Naming is hard :)

honzasp (commented):
This looks ok, but I assume that this is mostly a proof of concept: in particular, we must not guarantee backwards compatibility with this API, because we will need to resolve at least these things:

- Support other messaging/pubsub platforms
- Subscription to topics described by code, not only by files
- More ergonomic handling of events (we should not simply dump raw byte buffers on frontend devs :))
- Delivery of events, how to guarantee at-most-once or at-least-once semantics.

cli/src/project.rs
penberg: @honzasp I think "events" is easier to grasp than "consumers" so I am keeping it.

However, if we switch to `routes`, then `topics` makes perfect sense.

penberg:
@honzasp Yes, this is indeed an experimental feature. Thanks for reminding me about that -- updated the documentation to reflect that.

honzasp (approved):

PR #1640, opened 2022-08-12
native date types

glommer:
Works on my tests, probably doesn't work very well with nested types.


ricardodalarme:
this going to be really useful!!!

PR #1641, opened 2022-08-12
fully qualified test names

MarinPostma:
fully qualify test names to allow precise filtering.


honzasp:
Hmm, if I understand it correctly, it produces the full module path, so the test name would be something like `integration_tests::rust_tests::bad_filter::test`; would it be possible to skip everything up to the `rust_tests` part, without getting too deep in some macro wizardry? Or do you think that the full name is ok?

MarinPostma:
uhm... I don't think so. This is also how the rust test harness would name them btw

honzasp (approved):

PR #1645, opened 2022-08-14
Implement token authentication

dekimir:
A new kind of policy lets you forbid endpoint access without a particular HTTP header that contains the correct secret value.

### DIFF
--- a/cli/tests/integration_tests/framework.rs
+++ b/cli/tests/integration_tests/framework.rs
@@ -1,6 +1,7 @@
 use crate::database::Database;
 use anyhow::Result;
 use bytes::{Bytes, BytesMut};
+use reqwest::header::HeaderMap;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::process::{ExitStatus, Stdio};
@@ -316,6 +317,10 @@ impl Chisel {
         self.exec("wait", &[]).await
     }
 
+    pub async fn restart(&self) -> Result<ProcessOutput, ProcessOutput> {
+        self.exec("restart", &[]).await
+    }
+
     /// Writes given `text` (probably code) into a file on given relative `path`
     /// in ChiselStrike project.
     pub fn write(&self, path: &str, text: &str) {
@@ -353,7 +358,13 @@ impl Chisel {
     /// Sends a HTTP request to a relative `url` on the running `chiseld`, using the given request
     /// `method` and `body`. Does not check that the response is successful. Panics if there was an error while
     /// handling the request.
-    pub async fn request<B>(&self, method: reqwest::Method, url: &str, body: B) -> reqwest::Response
+    pub async fn request_with_headers<B>(
+        &self,
+        method: reqwest::Method,
+        url: &str,
+        body: B,
+        headers: HeaderMap,
+    ) -> reqwest::Response
     where
         B: Into<reqwest::Body>,
     {
@@ -364,10 +375,19 @@ impl Chisel {
         self.client
             .request(method.clone(), full_url)
             .body(body)
+            .headers(headers)
             .timeout(core::time::Duration::from_secs(5))
             .send()
             .await
             .unwrap_or_else(|e| panic!("HTTP error in {} {}: {}", method, url, e))
+    }
+
+    pub async fn request<B>(&self, method: reqwest::Method, url: &str, body: B) -> reqwest::Response
+    where
+        B: Into<reqwest::Body>,
+    {
+        self.request_with_headers(method, url, body, HeaderMap::new())
+            .await
     }
 
     /// Same as `request()`, but reads the response status and body as bytes.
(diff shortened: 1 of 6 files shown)


glommer:
This looks good to me in general, and tests are good.
First time I read the test I had the impression you were adding the token in plain text in the policy file. I then saw you aren't, but I wonder if I wasn't thrown off scent by it being pointed to by "value".

Maybe a more descriptive attribute name could help ?

glommer:
ack, but tests fail

dekimir:
@glommer tests fixed, added per-verb control.

cli/tests/integration_tests/framework.rs
```diff
@@ -483,14 +487,15 @@ impl Chisel {
...
     /// Gets an `url` of a running ChielStrike service.
-    pub async fn get(&self, url: &str) -> Result<reqwest::Response> {
-        let url = url::Url::parse(&format!("http://{}", self.config.public_address))
-            .unwrap()
-            .join(url)
-            .unwrap();
+    pub async fn get(
+        &self,
+        url: &str,
+        headers: reqwest::header::HeaderMap,
```
honzasp, line 493: I think we should add a new method that supports request headers; most tests don't need to pass any headers, and this forces them to pass empty headers every time.

honzasp: (This applies to all other methods.)

cli/tests/integration_tests/rust_tests/test_token_auth.rs
```diff
@@ -0,0 +1,147 @@
+use crate::framework::TestConfig;
+use crate::framework::{IntegrationTest, OpMode};
+use reqwest::{header::HeaderMap, StatusCode};
+use serde_json::json;
+
+fn header(name: &'static str, value: &str) -> HeaderMap {
+    let mut headers = HeaderMap::new();
+    headers.insert(name, value.parse().unwrap());
+    headers
+}
```
honzasp, line 10: We should perhaps consider adding these generic helper functions to `framework.rs`, so that they are available to all tests.

server/src/policies.rs
```diff
@@ -71,10 +73,78 @@ impl UserAuthorization {
...
+}
+
+/// Describes a header that a request must include.
+#[derive(Clone, Default, Debug)]
+struct RequiredHeader {
+    header_name: String,
+    /// Names a secret (see secrets.rs) whose value must match the header value.
+    secret_name: String,
+    /// HTTP verbs to which this requirement applies.  If absent, apply to all verbs.
+    verbs: Option<Vec<hyper::Method>>,
```
honzasp, line 140: Maybe we should use "method" instead of "verb", to be consistent with the rest of the code?

honzasp (commented):
This looks reasonable, but what are our intentions with this feature? Authentication based on a static global password is not going to be considered very secure in 2022 and it will not work at all when the backend is accessed from the client, so we probably should not market this as a canonical way to secure our backends.

cli/tests/integration_tests/rust_tests/test_token_auth.rs
```diff
@@ -0,0 +1,134 @@
...
+   mandatory_header: { name: header33, value: TOKEN33 }
+ - path: /
+   mandatory_header: { name: foo, value: BAR }"##,
+    );
+    chisel
+        .apply()
+        .await
+        .expect_err("Didn't catch repeat path")
+        .stderr()
+        .peek("Repeated path in header authorization");
```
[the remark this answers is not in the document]
dekimir: Isn't `read` slower?

server/src/policies.rs
dekimir: "Method" is quite ambiguous, though.  We'd have to call it "HTTP methods" to disambiguate, which is inelegant.

cli/tests/integration_tests/rust_tests/test_token_auth.rs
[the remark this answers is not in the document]
dekimir: I wasn't.  I read the source.  I can't imagine `.find` being faster than `.contains`.

glommer:
> so we probably should not market this as a canonical way to secure our backends.

it doesn't have to be the canonical way, just a way, that is good enough for some simple applications.

These tokens can expire btw. You expire them by resetting your .env =p

cli/tests/integration_tests/framework.rs
dekimir: Done.

cli/tests/integration_tests/rust_tests/test_array.rs
```diff
@@ -60,15 +60,15 @@ pub async fn test_array(config: TestConfig) {
         .expect("failed to store data");
 
     {
-        let data = chisel.get_json("/dev/get").await;
+        let data = chisel.get_json("/dev/get", HeaderMap::new()).await;
```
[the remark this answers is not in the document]
dekimir: Done 7d35de81

cli/tests/integration_tests/framework.rs
```diff
@@ -547,6 +547,7 @@ fn chiseld() -> String {
 
 async fn setup_chiseld(config: &TestConfig) -> Result<(Database, GuardedChild, Chisel)> {
     let tmp_dir = Arc::new(TempDir::new("chiseld_test")?);
+    let path = tmp_dir.path().to_owned();
```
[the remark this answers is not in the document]
dekimir: Done 75523756

dekimir:
Finally rebased. :sweat: All feedback has been addressed.

cli/tests/integration_tests/rust_tests/test_token_auth.rs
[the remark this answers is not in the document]
dekimir: Let's not bikeshed this.  I can use `read()` as a default in the future.

server/src/policies.rs
honzasp: It might be, but we use "method" everywhere else (both internally and in user-facing TypeScript APIs), so I guess it is even more inelegant to start using "verb" in policies.

Also, "method" is the correct technical term from the HTTP specifications, it is used almost universally in libraries, frameworks and APIs that deal with HTTP, and it is searched 10x more often than "verb" according to Google Trends [0].

[0] https://trends.google.com/trends/explore?date=all&q=http%20method,http%20verb

PR #1652, opened 2022-08-17
Policies: rename verbs to methods.

BearLemma:
Renaming verbs to methods as that seems more customary. (https://stackoverflow.com/a/49908515/1474847)

### DIFF
--- a/api/src/chisel.ts
+++ b/api/src/chisel.ts
@@ -1393,7 +1393,7 @@ export type CRUDMethodSignature<
 ) => Promise<Response>;
 
 /**
- * A dictionary mapping HTTP verbs into corresponding REST methods that process a Request and return a Response.
+ * A dictionary mapping HTTP methods into corresponding REST methods that process a Request and return a Response.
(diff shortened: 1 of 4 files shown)


penberg (approved):
If @dekimir doesn't like this, we can revert.

honzasp (approved):

glommer (changes requested):
this can no longer go in.
(and it is a great example of why we should be careful in the future with merges)

We now have a release, 0.12, in which users make reference to `verbs` in yaml and may be already using it.

We have to either live with "verbs", or change it so that both "verbs" and "methods" are accepted in the yaml file.

penberg (approved):

honzasp:
> (and it is a great example of why we should be careful in the future with merges)

Well, both Jenda and me had brought up this issue in that PR multiple times, Dejan explicitly decided to keep using "verb".

But I think it should be easy to retain backwards-compatibility by supporting both variants in YAML.

PR #1656, opened 2022-08-18
docs: Switch to JSON-based streaming example

penberg:
The StringDecoder thing is just ugly and nobody will use that in
real-life code. Switch to equally ugly, yet useful, example with JSON.

### DIFF
--- a/docusaurus/docs/Intro/streaming.md
+++ b/docusaurus/docs/Intro/streaming.md
@@ -11,11 +11,12 @@ For example, to consume events from a `hello` topic, create a `events/hello.ts` 
 ```typescript title="my-backend/events/hello.ts"
 import { ChiselEvent } from "@chiselstrike/api";
 
+function toJSON(buffer: ArrayBuffer) {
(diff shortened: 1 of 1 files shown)


docusaurus/docs/Intro/streaming.md
```diff
@@ -11,11 +11,12 @@ For example, to consume events from a `hello` topic, create a `events/hello.ts`
 ```typescript title="my-backend/events/hello.ts"
 import { ChiselEvent } from "@chiselstrike/api";
 
+function toJSON(buffer: ArrayBuffer) {
```
glommer, line 14: Pekka: nothing in this function takes anything specific to the buffer into consideration.
Can't we librarify it?

penberg: @glommer We can, but I don't know if it's the API we want to do. In any case: seems unrelated to this pull request, which just fixes up the docs to something nicer (even if not perfect).

glommer (approved):

PR #1667, opened 2022-08-19
Use `pub` instead of `pub(crate)` in server (from #1618)

honzasp:
We don't have to use `pub(crate)` everywhere, it is sufficient if we apply `pub(crate)` only at the crate root and use just `pub` in modules.

In particular, this does not affect warnings about dead code.


MarinPostma (dismissed):

honzasp:
I had to apply rustfmt, can you please reapprove? :)

PR #1670, opened 2022-08-19
Use "begin transaction", not "start transaction" (from #1618)

honzasp:
In SQL, the statement is `BEGIN TRANSACTION`, there is no need to introduce custom terminology.


penberg:
Looks good, will approve when tests pass

penberg (approved):

PR #1679, opened 2022-08-19
Don't use unnecessary polymorphism (from #1618)

honzasp:
`&str` is isomorphic to `T: AsRef<str>`, but does not generate redundant monomorphisations of the same function.

### DIFF
--- a/server/src/api.rs
+++ b/server/src/api.rs
@@ -16,8 +16,6 @@ use std::convert::Infallible;
 use std::convert::TryFrom;
 use std::io::Cursor;
 use std::net::ToSocketAddrs;
-use std::path::Path;
-use std::path::PathBuf;
 use std::pin::Pin;
 use std::rc::Rc;
 use std::sync::{Arc, Mutex};
@@ -123,15 +121,8 @@ pub struct ApiInfo {
 }
 
 impl ApiInfo {
-    pub fn new<N, T>(name: N, tag: T) -> Self
-    where
-        N: ToString,
-        T: ToString,
-    {
-        Self {
-            name: name.to_string(),
-            tag: tag.to_string(),
-        }
+    pub fn new(name: String, tag: String) -> Self {
+        Self { name, tag }
     }
 
     pub fn all_routes() -> Self {
@@ -150,7 +141,7 @@ impl ApiInfo {
         }
     }
 }
-pub type ApiInfoMap = HashMap<PathBuf, ApiInfo>;
+pub type ApiInfoMap = HashMap<String, ApiInfo>;
 
 /// API service for Chisel server.
 pub struct ApiService {
@@ -159,7 +150,7 @@ pub struct ApiService {
     // with runtime checking, which is likely cheaper, but still this is safer and we don't
     // have to manually implement Send (which is unsafe).
     paths: Mutex<PrefixMap<RouteFn>>,
-    event_handlers: Mutex<HashMap<PathBuf, EventFn>>,
+    event_handlers: Mutex<HashMap<String, EventFn>>,
     info: Mutex<ApiInfoMap>,
 }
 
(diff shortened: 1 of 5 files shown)


glommer:
Honza, why?

do you have any indication that the rust compiler will generate worse code with those classes?
If the only caller is &str, I expect the compiler to be able to figure this out on its own.

there is a big advantage in using `AsRef<str>`, `ToString`, and friends: other callers that have other types don't have to convert explicitly, if there is an implicit conversion.

It is true that it is not *needed* when the only caller is of a certain type, but if it doesn't generate worse code, this is exactly the kind of stuff that computers should worry about, not humans.

MarinPostma:
@glommer, I may be wrong but I think the rationale here is that monomorphization slows down compile time? But then it's probably anecdotical in comparison to the amount of monomorphization that's already happening, so maybe @honzasp can explain why he thinks that's a necessary change.

honzasp:
The issue is not that the compiler generates worse code, the issue is that the compiler has to generate a copy of the function for every type that you happen to use. In these cases it is probably negligible, but I saw big functions with ~50 lines of code being monomorphised twice just to abstract away a `.to_string()`.

The costs of this are increased size of generated code, slower compilation, bigger files, more confusing debugging (one function in source corresponds to _N_ independent functions in the generated code). None of this is very important in this particular case, but it might if this pattern grows unchecked.

This is just a bad habit.

honzasp:
By this logic, we should replace all `&T` function parameters with `impl AsRef<T>`, which I think we all agree would be ridiculous.

MarinPostma:
> By this logic, we should replace all &T function parameters with impl AsRef<T>, which I think we all agree would be ridiculous.
This is a bad faith argument and is only true if impl AsRef<T> is isomorphic to &T, which is not true for `AsRef<Path>` the, is not isomorphic to `&Path` for instance.

glommer:
This is not just a bad faith argument. It is also adversarial for no reason, by arguing the extreme case to demonstrate the underlying rule. Code is written by humans (for now), so common sense plays a role. The extreme case doesn't matter.

Conversely you should ask yourself why, even in the presence of this pattern, you don't see things like `impl AsRef<T>` for every `T`. And the reason for that is that in practice, there are a couple of types for which deref to others very commonly. String, Path, etc. A function taking `AsRef<Path>` can take a PathBuf, a Path, a String, a &str, *and* it is very common in practice to have one or the other from calling contexts.

I am happy to merge this so life can move on, but this is a pattern that I happen to like (when used reasonably, for types like Path and String). I would argue that it is almost crucial for public APIs (which those are not, which is why I am ok bending), and I think you are optimizing for the wrong thing here.

glommer (approved):

honzasp:
> This is a bad faith argument

Not sure what you mean by this fashionable expression, but the technical term for this type of argument is _reductio ad absurdum_ ("if these arguments for `AsRef<str>` make sense, then we should be using `AsRef<T>` much more often, which is clearly absurd")

> is only true if impl AsRef is isomorphic to &T, which is not true for AsRef<Path> the, is not isomorphic to &Path for instance.

`impl AsRef<Path>` is isomorphic to `&Path`, because the only meaningful thing that you can do with such value is to convert it to `&Path`.

> Conversely you should ask yourself why, even in the presence of this pattern, you don't see things like `impl AsRef<T>` for every `T`.

Yes, this is exactly what I'm asking :) One could use exactly the same arguments for using, say, `impl AsRef<Policies>` instead of `&Policies`, because it makes it convenient to pass `&Policies`, `Arc<Policies>` or `Box<Policies>`. We don't do this, and perform the conversion in the caller (which in most cases is just a single `&`) instead of generating a new copy of the callee for every combination of types that we happen to use. What is so different about `str` and `Path`?

> I would argue that it is almost crucial for public APIs

Maybe, but the pattern here is that the polymorphic public function simply converts the arguments and calls the monomorphic function that performs the actual work, precisely to avoid needless copies of functions.
https://doc.rust-lang.org/src/std/fs.rs.html#2174-2176

There is even a crate that abstracts this pattern
https://github.com/llogiq/momo

> I think you are optimizing for the wrong thing here.

The `target/` directory on my computer takes 35 GB, so compilation size (and speed) _do_ bother me :) Again, any performace/code size/compilation speed implications are trivial in this PR, but they could quickly get meaningful if we started to use this pattern more heavily.

What I _am_ always optimizing for is clarity and simplicity. Which of these two snippets is more clear and simple?

```
// before
    pub fn new<N, T>(name: N, tag: T) -> Self
    where
        N: ToString,
        T: ToString,

// after
    pub fn new(name: String, tag: String) -> Self
```

(Note that this PR reduced the number of lines, so the code without `AsRef<>` is simpler _and_ shorter.)

PR #1681, opened 2022-08-19
Don't print "End point defined" from `chisel apply` (from #1618)

honzasp:
With the new routing API, we will be unable to print these messages, so let's remove them now.

### DIFF
--- a/cli/src/cmd/apply.rs
+++ b/cli/src/cmd/apply.rs
@@ -178,21 +178,11 @@ pub(crate) async fn apply(
 
     let msg = execute!(client.apply(tonic::Request::new(req)).await);
 
-    for ty in msg.types {
-        println!("Model defined: {}", ty);
-    }
-
-    for end in msg.endpoints {
-        println!("End point defined: {}", end);
-    }
-
-    for end in msg.event_handlers {
-        println!("Event handler defined: {}", end);
-    }
-
-    for lbl in msg.labels {
-        println!("Policy defined for label {}", lbl);
-    }
+    println!("Code was applied to the ChiselStrike server. It contained:");
+    println!("  - models: {}", msg.types.len());
+    println!("  - endpoints: {}", msg.endpoints.len());
+    println!("  - event handlers: {}", msg.event_handlers.len());
+    println!("  - labels: {}", msg.labels.len());
 
     Ok(())
 }
--- a/cli/tests/integration_tests/lit_tests/access-origin.deno
+++ b/cli/tests/integration_tests/lit_tests/access-origin.deno
@@ -16,7 +16,7 @@ EOF
 cd "$TEMPDIR"
 $CHISEL apply
 
-# CHECK: End point defined: /dev/foo
+# CHECK: Code was applied
 
 $CURL -o - $CHISELD_HOST/dev/foo
 
--- a/cli/tests/integration_tests/lit_tests/aggregations.deno
+++ b/cli/tests/integration_tests/lit_tests/aggregations.deno
@@ -86,8 +86,7 @@ export default async function chisel(req: Request) {
 EOF
 
 $CHISEL apply
-# CHECK: Model defined: Person
-# CHECK: End point defined: /dev/compute_aggregations
+# CHECK: Code was applied
 
 $CURL -X POST -o - $CHISELD_HOST/dev/store
 # CHECK: Ok
--- a/cli/tests/integration_tests/lit_tests/api-policy.deno
+++ b/cli/tests/integration_tests/lit_tests/api-policy.deno
@@ -27,10 +27,7 @@ EOF
 cd "$TEMPDIR"
 $CHISEL apply
 
-# CHECK: Model defined: Person
-# CHECK: End point defined: /dev/query
-# CHECK: End point defined: /dev/store
-# CHECK: Policy defined for label pii
+# CHECK: Code was applied
 
 $CURL --data '{
     "first_name":"Glauber",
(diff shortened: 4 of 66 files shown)


penberg (changes requested):
We need to print something as feedback to `chisel apply`, but more importantly, `chisel dev`.

What _can_ we print with new routing?

glommer:
FWIW Pekka, I don't oppose not printing every route. In fact, we have heard complains that the fact that we print all endpoints all the time is confusing, and the code to check whether or not an endpoint changed so we could print just the changes is nasty. (I tried once, and it was easily too complex). So I don't oppose this change, but we should replace it with something indeed.

Remove the policy ones as well while we're at it. Maybe all we need is an apply-wide message.

honzasp:
I agree, printing _something_ is also important for the lit tests, where the `End point defined` checks were used as "barriers" for matching the `CHECK`s to the output.

penberg:
The concrete issue is that with these changes `chisel dev` output becomes even more confusing.

With a "hello world" endpoint, startup looks as follows:

```
penberg@vonneumann my-backend % ../target/debug/chisel dev
🚀 Thank you for your interest in the ChiselStrike beta! 🚀

⚠️  This software is for evaluation purposes only. Do not use it in production. ⚠️

📚 Docs:    https://docs.chiselstrike.com
💬 Discord: https://discord.gg/4B5D7hYwub
📧 Email:   beta@chiselstrike.com

For any question, concerns, or early feedback, please contact us via email or Discord!

[2022-08-22T06:03:52Z] INFO - RPC is ready. URL: 127.0.0.1:50051
[2022-08-22T06:03:52Z] INFO - ChiselStrike is ready 🚀 - URL: http://localhost:8080
[2022-08-22T06:03:53Z] INFO - Restarting
[2022-08-22T06:03:53Z] INFO - RPC is ready. URL: 127.0.0.1:50051
[2022-08-22T06:03:53Z] INFO - ChiselStrike is ready 🚀 - URL: http://localhost:8080
```

If I now go and change the endpoint to have a syntax error, I see:

```
[2022-08-22T06:06:23Z] INFO - Restarting
[2022-08-22T06:06:23Z] INFO - RPC is ready. URL: 127.0.0.1:50051
[2022-08-22T06:06:23Z] INFO - ChiselStrike is ready 🚀 - URL: http://localhost:8080
Error: Parse failed: error: Expected ';', got 'xxx'
 --> <anon>:8:46
  |
8 |     return await req.text() || "hello world" xxx;
  |                                              ^^^


[2022-08-22T06:06:23Z] INFO - Restarting
[2022-08-22T06:06:23Z] INFO - RPC is ready. URL: 127.0.0.1:50051
[2022-08-22T06:06:23Z] INFO - ChiselStrike is ready 🚀 - URL: http://localhost:8080
could not import endpoint code into the runtime

Caused by:
    Error: expected type `v8::data::Function`, got `v8::data::Value`
        at importEndpointsImpl (file:///worker.js:54:19)
        at async handleMsg (file:///worker.js:23:17)
```

but because the endpoint printout is dropped, there's no clue of what's going on. I am not claiming it was perfect before, but now it's even worse.

honzasp:
Yes, you are absolutely right, we need to print something that shows that `chisel apply` was successful.

Another issue is that `chisel dev` prints the output from `chiseld` which says `ChiselStrike is ready 🚀 - URL: http://localhost:8080`: this shows that `chiseld` started successfully, but _not_ that the code was successfully applied.

penberg:
@honzasp One possible fix here is to move that `ChiselStrike is ready` logging to `chisel dev` and `chisel start` and drop that completely from `chiseld`. We can then only print it once when we start up. We could then just do something like `Reloaded <list of files>` thing in `chisel apply` once the apply RPC completes?

honzasp:
I rewrote the PR as follows:
- `chisel apply` now prints just a single message that says "Code was applied to the ChiselStrike server"
- The colorful message from `chiseld` now says "ChiselStrike *server* is ready 🚀"

Is this OK for now?

penberg:
@honzasp I think it's still a regression from the developer experience we have. Event handlers and events are now handled differently as well.

We know the source file names at `chisel apply` time, no? Can we make it print something like after a successful apply?

```
Applied modules:
  - endpoints/foo.ts
  - events/bar.ts
```

honzasp:
Yes, we could print that, but for any non-trivial project this will just spam your console with tens of files for every refresh. For example, the dogfood backend is made from >100 files, we probably should not print them every time.

As Glauber commented previously, we could eventually print just the files that were changed, but this is unfortunately impossible at this moment. I suggest that we open an issue and fix it later, after `chisel` is able to resolve modules.

penberg:
@honzasp So we can do log rate limiting on the files. For example, print first 10 files, but if there are more, just print out the directory or something like:

```
Applied modules:
  - endpoints/foo.ts
  - events/bar.ts
  - [20 files omitted from output]
```

We don't have to aim for perfection, we just need some usable feedback for the developer using `chisel dev`.

honzasp:
Ok, so what information do we want to display to the developer when we reload the code, beside the fact that we applied the changes successfully? Perhaps the identifier of the version?

honzasp:
> @honzasp _My_ preference is still listing the files that were reloaded (with rate limiting) because as we move towards of applying things incrementally, it will just work.
> 
> What @glommer seems to be suggesting seems to be something like:
> 
> ```
> Applied <number> modules from the `endpoints` directory.
> ```

Ok, that makes sense, but unfortunately we are unable to print this information now :( We don't know which modules are loaded, for example. But we could print the number of files from `endpoints/` (or `routes/`) that we have read, so I can perhaps do this in this PR and leave any extensions for future PRs?

penberg:
@honzasp Yeah, I think number of files is good enough approximation for what Glauber was suggesting.

penberg (approved):

PR #1685, opened 2022-08-22
Containerize Github actions.

BearLemma:
This PR containerizes out pipeline builds to achieve better isolation from Runner's filesystem, stray processes etc.

It's currently based on our Docker image stored in chiselstrike's GH container registry: `ghcr.io/chiselstrike/rust-build-image:v1.0` which was built using `.github/build.Dockerfile`.
The reason why we need to have custom container is that we need to run the tests without root access to test some properties of file access and manipulation. That however doesn't allow us to install the required packages. (I got `sudo not found` error). So we build a container with dependencies and run in it without root. 
 
I wanted to automatically build the image as a pre-step to our build process, but I'm unable to make it push to our registry so I've deferred it to another PR which doesn't even need to go in as I don't expect the dependencies to change very often and we can just push manually...

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,27 +20,14 @@ env:
 jobs:
   build:
     runs-on: self-hosted
+    container:
+      image: ghcr.io/chiselstrike/rust-build-image:v1.0
+      options: --user 1000
     steps:
-    - name: Kill stray chiseld processes
-      run: "killall -9 -q chiseld || true"
     - name: checkout repo
-      uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -99,6 +99,7 @@ jobs:
 
   postgres-test:
     runs-on: self-hosted
+    container: ubuntu:22.10
```
penberg, line 102: Don't you need to containerize that other job too that runs tests with SQLite?

penberg (approved):

PR #1686, opened 2022-08-22
Remove the word "endpoint" from public interfaces (from #1618)

honzasp:
PR https://github.com/chiselstrike/chiselstrike/pull/1618 replaces the concept of "endpoint" with the concept of a "route", so let's change this now, before the main PR lands.

Notably, this changes the default directory "endpoints/" to "routes/". Unfortunately, even though the directory appears to be configurable in "Chisel.toml", it is in fact hard-coded in several places, so we have to hack around this to retain backward compatibility.

This change also affects some error messages, which are brought in alignment with https://github.com/chiselstrike/chiselstrike/pull/1618 by this commit.

### DIFF
--- a/api/src/worker.ts
+++ b/api/src/worker.ts
@@ -91,7 +91,7 @@ async function importEndpointsImpl(
         // Modules are never unloaded, so we need to create an unique
         // path. This will not be a problem once we publish the entire app
         // at once, since then we can create a new isolate for it.
-        const url = `file:///${apiVersion}/endpoints${path}`;
+        const url = `file:///${apiVersion}/routes${path}`;
         const mod = await import(url);
         const handler = mod.default;
         if (typeof handler !== "function") {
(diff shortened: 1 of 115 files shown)


penberg (changes requested):
For this to be merged, you also need to update the documentation, no?

I am not convinced we should do the `routes` directory change, because it's a lot of churn, but it's not obvious what the gains are.

honzasp:
You are right, I will update the documentation.

#1618 completely eliminates any concept of "endpoint", so it does not make sense to keep using "endpoints/" as a directory name. It is as if we used the name "rules/" for the directory that contains policies or "models/" for the directory that contains entities.

This PR will bring the repo in a temporary weird state: "endpoints" will still exist in the code, but the external interface will now use "routes". Hopefully, this weirdness will be resolved by merging #1618 soon.

honzasp:
I updated the tutorial to use the "routes" directory, and also replaced the word "endpoint" where it was natural. We can still use the word "endpoint" as an informal synonym for a given route and its handler (such as `GET /comment/:id`).

Btw, the docs will need some copy-editing; I fixed some typos when I saw them, but I didn't do any large edits. We will need to update the docs if the `RouteMap` API lands in #1618, so we can also use the opportunity to refactor the docs a bit.

penberg (approved):

PR #1687, opened 2022-08-22
Remove the `Chisel` global variable (from #1618)

honzasp:
In modern JavaScript, it is usually better to use imports rather than a global variable. For example, to type-check code that uses the global variable, we must feed `tsc` with an extra file that specifies the type of the variable. Getting rid of the `Chisel` variable will simplify our code in the fugure.


penberg:
@honzasp Looks good to me! Will let others also review this too because I wonder why we didn't get rid of it before.

penberg (approved):

PR #1688, opened 2022-08-22
Split the TypeScript code in `api/` into multiple modules (from #1618)

honzasp:
Split the large `chisel.ts` into multiple files that are imported from `api.ts`.

This requires some ad-hoc changes to `endpoint_tsc`; hopefully, this code will go away in foreseeable future, so I didn't bother refactoring this code too much.

There is one issue that had to be resolved: the files in `api/src/*.ts` import each other using full filenames (`import from './request.ts'`), but vanilla `tsc` requires imports without extensions. To support typechecking of Chiselstrike projects with vanilla `tsc`, I resorted to a hack and add an extra ".d.ts" extension to files in the npm package, i.e. "request.ts" is stored as "lib/request.ts.d.ts" in the package.

This is a spin-off from #1618, it depends on #1687.


honzasp:
Marked as draft, because this includes an unmerged commit from #1687. Otherwise, the PR should be ready.

penberg:
@honzasp Is there any user-visible change with imports when we merge this? IOW, will importing `@chiselstrike/api` have all the things needed?

honzasp:
> Is there any user-visible change with imports when we merge this? IOW, will importing `@chiselstrike/api` have all the things needed?

Yes, everything should still work.

penberg (approved):

PR #1697, opened 2022-08-24
Tutorial updates

penberg:


docusaurus/docs/Intro/first.md
```diff
@@ -92,10 +92,8 @@ To handle HTTP requests on "/dev/comments", we create a TypeScript file
 in the `my-backend/routes` directory.  Here is one:
 
 ```typescript title="my-backend/routes/comments.ts"
-import { responseFromJson } from "@chiselstrike/api"
-
-export default function (_req) {
-    return responseFromJson("Comments go here!");
+export default async function (req: Request) {
+    return "Comments go here!";
 }
```
honzasp, line 97: Note that this is no longer equivalent due to #1673, the response will simply contain the string verbatim. This is probably the correct behavior, but we should update the "and should see the following output:" section a few lines below this to remove the quotes?

PR #1699, opened 2022-08-25
Tests: Replace chisel restart by manual process restarting.

BearLemma:
Prepares ground for removal of `chisel restart` command.

### DIFF
--- a/cli/tests/integration_tests/framework.rs
+++ b/cli/tests/integration_tests/framework.rs
@@ -23,23 +23,24 @@ pub mod prelude {
 
 pub struct GuardedChild {
     child: tokio::process::Child,
+    command: tokio::process::Command,
     pub stdout: AsyncTestableOutput,
     pub stderr: AsyncTestableOutput,
 }
 
 impl GuardedChild {
-    pub fn new(command: &mut tokio::process::Command) -> Self {
-        let mut child = command
-            .stdout(Stdio::piped())
+    pub fn new(mut cmd: tokio::process::Command) -> Self {
+        cmd.stdout(Stdio::piped())
             .stderr(Stdio::piped())
-            .kill_on_drop(true)
-            .spawn()
-            .expect("failed to spawn GuardedChild");
+            .kill_on_drop(true);
+
+        let mut child = cmd.spawn().expect("failed to spawn GuardedChild");
 
         let stdout = child.stdout.take().unwrap();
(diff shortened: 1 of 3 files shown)


cli/tests/integration_tests/framework.rs
```diff
@@ -23,23 +23,20 @@ pub mod prelude {
...
 
 impl GuardedChild {
-    pub fn new(command: &mut tokio::process::Command) -> Self {
-        let mut child = command
-            .stdout(Stdio::piped())
-            .stderr(Stdio::piped())
-            .kill_on_drop(true)
-            .spawn()
-            .expect("failed to spawn GuardedChild");
+    pub fn new(cmd_factory: Box<dyn Fn() -> tokio::process::Command + Send>) -> Self {
```
honzasp, line 32: Could we take simply `tokio::process::Command` here? We could call the `.stdout(Stdio::piped())` method and the rest from `spawn_chiseld()` here just once, and then simply use `spawn()` on the `Command` multiple times (it takes `&mut self`).

cli/tests/integration_tests/framework.rs
```diff
@@ -57,6 +54,30 @@ impl GuardedChild {
...
         self.stderr.show().await;
     }
+
+    pub async fn restart(&mut self) {
+        use nix::sys::signal;
+        use nix::unistd::Pid;
+
+        let pid = Pid::from_raw(self.child.id().unwrap().try_into().unwrap());
+        signal::kill(pid, signal::Signal::SIGTERM).unwrap();
+        let r = tokio::time::timeout(Duration::from_secs(10), self.wait()).await;
```
honzasp, line 64: Maybe we should use a shorter timeout, like 1 second or less? No need to block the test if `chiseld` hangs...

cli/tests/integration_tests/framework.rs
```diff
@@ -57,6 +54,30 @@ impl GuardedChild {
...
+
+        if r.is_err() {
+            eprintln!("failed to successfully terminate child process. Going for the kill...");
```
honzasp, line 67: This print may disturb test outputs (it will be hard to assign the print to the test that caused it, for example), do you think it would be OK to kill the process silently?

honzasp (commented):
I like this! :) I have just a few minor comments on the implementation.

honzasp (approved):

PR #1705, opened 2022-08-25
api: Switch ChiselEvent to use Blob type

penberg:
We need to persist key and value to a database to make event publishing
transactional. To do that, let's switch the `ChiselEvent` to use the
`Blob` type, which is a nicer type for an ORM.

As a bonus, it becomes easier to consume events as JSON as illustrated
in the docs.


penberg:
@honzasp It's a _nicer_ type, because entity definitions read out better with `Blob` (it's a self-explanatory type).

honzasp:
Hmm, I understand that `Blob` looks better than `ArrayBuffer`, but there might be other reasons for using `ArrayBuffer`. For example, a `Blob` does not have to be a memory buffer (it can be a file, for example), so we would need to call the async method `Blob.arrayBuffer()` to convert it to an `ArrayBuffer` before we pass it to Rust and store in the database. This is definitely possible, but it might add unnecessary complications.

penberg:
@honzasp The `ChiselEntity.save()` method is already `async` converting to `ArrayBuffer` is not a problem.

But thinking about this, I am more convinced that `Blob` is the right type because it is _immutable_, which is the right thing for `Event` whereas `ArrayBuffer` can be manipulated via a view.

honzasp:
> The `ChiselEntity.save()` method is already `async` converting to `ArrayBuffer` is not a problem.

Yes, we will have to do the translation `Blob` -> `ArrayBuffer` in JavaScript, but we don't know the schema in JS, so the only thing that we could do is to walk through the object (like `deepMerge`) and replace all `Blob`s with `ArrayBuffer`s.

A similar problem also appears in the other direction, when returning the entity data from Rust: `ArrayBuffer` is built into JavaScript, but `Blob` is defined somewhere in Deno, so it will be much less convenient to create an instance of `Blob` from Rust.

Lastly, if we use `Blob` in entities, we must also store the MIME type of the blob with the data, so we must in fact replace the `Blob` with something like `{ contentType: string, data: ArrayBuffer }` before passing it to Rust.

honzasp:
This is related to issue #1739

7flash:
In Deno Blobs are always stored in memory

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix

penberg:
@honzasp I think `Blob` is the right type to use for this. So merge ping?

glommer (approved):

Issue #1655, opened 2022-08-18
Lack of documentation on entity relationships


penberg:
There's currently no documentation on how to define entity relationships.

PR #1711, opened 2022-08-26
Write docs on entity relationships.

BearLemma:
Closes https://github.com/chiselstrike/chiselstrike/issues/1655. 
As you can see, the docs I wrote are not exhaustive, but I didn't feel like putting my heart into it as we are likely to remove eagerly loaded relationships.

### DIFF
--- a/docusaurus/docs/Intro/data-access.md
+++ b/docusaurus/docs/Intro/data-access.md
@@ -252,10 +252,9 @@ export class BlogComment extends ChiselEntity {
 }
 ```
 
-You can observe that the `author` field is of type `User` which is another entity. This is a way we can define a relationship between entities.
+You can observe that the `author` field is of type `User`, which is another entity. This is a way we can define a relationship between entities.
 
-Entity fields are eagerly loaded. This means that when you load a `BlogComment` instance, the author field entity is loaded with it. Same goes for saving - when you save `BlogComment` with an author which doesn't yet exist, a new `User` instance will be created in the database and reference to it will be associated with stored `BlogComment`. If it's an already existing `User`, we will store a reference in this instance update the existing `User` if there are any changes.
-In other words, we will always upsert related entities.
+Entity fields are eagerly loaded. This means that when you load a `BlogComment` instance, the author field entity is loaded with it. Same goes for saving -- when you save a `BlogComment` with an author, a `User` instance will be upserted in the database and reference to it will be associated with the stored `BlogComment`. (So if this `User` already existed in the database, it will be updated with the value provided.)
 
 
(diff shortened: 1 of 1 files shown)


penberg (commented):
Looks good to me. Maybe @dekimir could also review this?

docusaurus/docs/Intro/data-access.md
```diff
@@ -237,6 +238,24 @@ export default async function (req: Request) {
...
+
+export class BlogComment extends ChiselEntity {
+    content: string = "";
+    author: User = new User();
+}
+```
+
+You can observe that the `author` field is of type `User` which is another entity. This is a way we can define a relationship between entities.
+
+Entity fields are eagerly loaded. This means that when you load a `BlogComment` instance, the author field entity is loaded with it. Same goes for saving - when you save `BlogComment` with an author which doesn't yet exist, a new `User` instance will be created in the database and reference to it will be associated with stored `BlogComment`. If it's an already existing `User`, we will store just a reference to this instance not creating a duplicate `User`.
```
dekimir, line 257: This doesn't seem quite correct.  Reading the code, it seems to me like we'll update the already existing `User` with the new values:
https://github.com/chiselstrike/chiselstrike/blob/796ef59db6e98d5efb57c6fa9afb55fd41492ffa/server/src/datastore/engine.rs#L789

dekimir (commented):
A to-do for me, once this lands: explain how the [similar example](https://github.com/chiselstrike/chiselstrike/blob/796ef59db6e98d5efb57c6fa9afb55fd41492ffa/docusaurus/docs/InDepth/login.md?plain=1#L42) with AuthUser behaves differently.

docusaurus/docs/Intro/data-access.md
```diff
@@ -237,6 +238,25 @@ export default async function (req: Request) {
...
+
+export class BlogComment extends ChiselEntity {
+    content: string = "";
+    author: User = new User();
+}
+```
+
+You can observe that the `author` field is of type `User`, which is another entity. This is a way we can define a relationship between entities.
+
+Entity fields are eagerly loaded. This means that when you load a `BlogComment` instance, the author field entity is loaded with it. Same goes for saving -- when you save a `BlogComment` with an author, a `User` instance will be upserted in the database and reference to it will be associated with the stored `BlogComment`. (So if this `User` already existed in the database, it will be updated with the value provided.)
```
dekimir, line 257: TODO: be clearer about the `author.id` field -- what happens if it's not provided, if it is provided and matches, if it is provided and doesn't match.

dekimir (approved):

PR #1714, opened 2022-08-26
Rewrite evolve-type.deno test to rust

BearLemma:


cli/tests/integration_tests/framework.rs
```diff
@@ -646,3 +654,71 @@ pub fn header(name: &'static str, value: &str) -> HeaderMap {
     headers.insert(name, value.parse().unwrap());
     headers
 }
+
+pub fn json_is_subset<V1, V2>(val: V1, subset: V2) -> Result<()>
+where
+    V1: Borrow<serde_json::Value>,
+    V2: Borrow<serde_json::Value>,
```
honzasp, line 661: For this large function, I think it would be preferable to use simply a reference to `serde_json::Value`, instead of generating up to 4 same copies for every combination of `Value` and `&Value`.

cli/tests/integration_tests/rust_tests/model_defaults.rs
```diff
@@ -0,0 +1,68 @@
+
+use serde_json::json;
+
+use crate::framework::prelude::*;
+use crate::framework::{json_is_subset, Chisel};
+
+fn write_crud_endpoint(chisel: &Chisel) {
```
honzasp, line 7: Perhaps we should use "route" instead of "endpoint" :)

cli/tests/integration_tests/rust_tests/model_evolve.rs
```diff
@@ -0,0 +1,273 @@
...
+    );
+    c.chisel.apply_ok().await;
+
+    let r = c.chisel.get_json("/dev/evolving").await;
+    json_is_subset(&r, json!({
+        "results": [{"a": "Heracles"}],
+    })).unwrap();
+    json_is_subset(r, json!({
+        "results": [{"a": "Heracles", "b": "with_default"}],
+    })).expect_err("Evolved type must not contain removed field b");
```
honzasp, line 93: I would make this check stronger by directly checking that there is no field `"b"` (the current check would also pass if `b` was a number or if `a` was removed, for example)

cli/tests/integration_tests/rust_tests/model_evolve.rs
```diff
@@ -0,0 +1,273 @@
...
+        export default async function chisel(req: Request) {
+            const filtered = Evolving.cursor()
+                .filter((e: Evolving) => {
+                    return e.a == "xx";
+                });
+            const results = (await filtered.toArray()).map(p => p.a);
+            return new Response("[" + results.join(", ") + "]");
+        }"##,
+    );
+    c.chisel.apply_ok().await;
```
honzasp, line 255: Can we somehow verify that an index is created, to make sure that this test actually tests something useful?

honzasp (commented):
This is great! :) Just a few nitpicks, really.

honzasp (approved):

PR #1721, opened 2022-08-29
Defaults.deno to Rust

BearLemma:
During the rewrite, I found that filtering over defaulted bools and numbers doesn't work. Fixed that as well. The tests I wrote will uncover that bug if it occurs again.


cli/tests/integration_tests/framework.rs
```diff
@@ -336,6 +336,11 @@ impl Chisel {
         self.exec("wait", &[]).await
     }
 
+    /// Runs `chisel wait` awaiting the readiness of chiseld service
```
penberg, line 339: The comment seems wrong

penberg (dismissed):

penberg (approved):

PR #1727, opened 2022-08-30
Rewrite savewrong.deno test to Rust.

BearLemma:


cli/tests/integration_tests/framework.rs
```diff
@@ -550,6 +550,20 @@ impl Response {
...
+            "Unexpected text response for HTTP {} {}\nResponse status {}, body {:?}, expected {:?}",
+            self.method,
+            self.url,
+            self.status,
+            actual,
+            expected,
+        );
+        self
+    }
+
```
MarinPostma, line 566: uhm you probably want to make that a macro, so the error is shown a the actual call points, or return an error from the function, and panic in the context of the test.

MarinPostma: then I think that returning an error and letting the caller unwrap is a good option?

MarinPostma: just replace the assert with `anyhow::ensure!`

MarinPostma: ok ok, your realm your rules 😄

MarinPostma (approved):

cli/tests/integration_tests/framework.rs
honzasp: The point here is that you get all the information about the response (status, body, etc.) in the error message, so this is vastly superior to just `assert!(response.text().contains(expected))`. Also, this being a method, these calls can "fluently" chain.

honzasp: (We could perhaps to the same magic that `unwrap()` uses to print the location of the caller?)

cli/tests/integration_tests/rust_tests/auth.rs
```diff
@@ -75,3 +75,85 @@ pub async fn next_auth_crud(mut c: TestContext) {
...
+
+    c.chisel
+        .post("/dev/auth_users")
+        .json(json!({"name":"Foo", "email":"foo@t.co"}))
+        .send()
+        .await
+        .assert_status(500)
+        .assert_text_contains("Error: Cannot save into type AuthUser.");
+    c.chisel
+        .post("/dev/auth_users/abcd-1234-eeee-5678")
```
dekimir, line 100: This was a PUT originally.

PR #1728, opened 2022-08-30
Share ports atomic between lit tests.

BearLemma:
I've noticed that while we do increment the ports, they get recycled for lit tests between unoptimized/optimized runs.
This PR fixes that and also bumps the ports 2K higher.


glommer:
the port number are by now essentially superstition =p 
But fine for now, will approve it if passes.

glommer (approved):

PR #1729, opened 2022-08-30
headers.deno to rust

BearLemma:
Had to split into two commits - first is `cargo fmt` as it wasn't formatted for some reason.


penberg (approved):

cli/tests/integration_tests/framework.rs
```diff
@@ -485,8 +490,9 @@ impl RequestBuilder {
 pub struct Response {
     method: reqwest::Method,
     url: reqwest::Url,
-    status: reqwest::StatusCode,
-    body: Bytes,
+    pub headers: reqwest::header::HeaderMap,
+    pub status: reqwest::StatusCode,
+    pub body: Bytes,
```
honzasp, line 495: Hmm, the intent here was that the fields are private, because you should use the methods :)

In this case, you should've added something like `pub fn header(&self, name: &str) -> String` which asserts that the header is present and valid utf-8 and that it returns the value.

honzasp: I'll open a PR

cli/tests/integration_tests/rust_tests/policies_loggedin.rs
```diff
@@ -332,3 +334,94 @@ async fn transform_match_login_related_entities(c: TestContext) {
...
+        .get("/__chiselstrike/auth/users")
+        .header("ChiselAuth", "u")
+        .header("ChiselUID", "1234")
+        .send()
+        .await;
+    resp.assert_ok();
+    assert!(contains_one_chisel_uid_header(&resp.headers));
+
+    // Store auth user and save uid
+    let resp = c
```
dekimir, line 401: Why not use `store_user()` here?

PR #1730, opened 2022-08-30
unique.deno to Rust

BearLemma:


cli/tests/integration_tests/rust_tests/model.rs
```diff
@@ -61,3 +61,53 @@ pub async fn duplicate_fields(c: TestContext) {
...
+
+    c.chisel
+        .post_json_ok(
+            "/dev/posts",
+            json!({"relUrl": "post.html", "content": "Hello World"}),
+        )
+        .await;
+    c.chisel
+        .post("/dev/posts")
+        .json(json!({"relUrl": "post.html", "content": "Hello World"}))
```
honzasp, line 97: Perhaps we should use a different value of `"content"`, to make sure that the uniqueness constraint is applied only to the `relUrl` field (and not, for example, to the tuple `(relUrl, content)` or to `content`)

honzasp (commented):
Great! Just a very small comment to make the test stronger.

honzasp (approved):

PR #1731, opened 2022-08-30
Remove restart command.

BearLemma:
Depends on https://github.com/chiselstrike/chiselstrike/pull/1729 and https://github.com/chiselstrike/chiselstrike/pull/1730

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -91,8 +91,6 @@ enum Command {
     Start,
     /// Show ChiselStrike server status.
     Status,
-    /// Restart the running ChiselStrike server.
-    Restart,
     /// Wait for the ChiselStrike server to start.
     Wait,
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -321,13 +319,6 @@ async fn main() -> Result<()> {
             let response = execute!(client.get_status(request).await);
             println!("Server status is {}", response.message);
         }
-        Command::Restart => {
-            let msg = match restart(server_url).await {
```
honzasp, line 325: Maybe we should also remove the `restart()` function, the `RestartRequest` in gRPC, and the code that handles the `RestartRequest` in the server?

honzasp: Ah, I see, so I will remove it in #1618

honzasp (approved):

PR #1733, opened 2022-08-30
Typescript policies Pt.1: barebone code analysis

MarinPostma:
This pr integrates a first draft of the typescript policy compiler.

In it's current state, this pr doesn't do much. It intergrates the basic code analysis, adds policies to the typesystem entities, and basically sets ground for policies to actually become usefull. It also adds logic to hide the policies behind a flag. For now, this flag is always set to false, and as soon as this get semi-usable, we'll expose it to the cli.

Right now, all tests should pass, and I have decided to cut off a PR here, so the size of this PR does not grow too much.

## How it works

For each Entity in `models/x.ts` one can create a policy in `policies/x.ts`, that defined how an entity can be accessed. The policy is written as an object that is exported by the module, like so:

```typescript
// in policies/Person.ts

export default {
    read: (person: Person, ctx: RequestContext) => {
    	// do something
    },
    write: (person: Person, ctx: RequestContext) => {
        // do something
    },
    transform: (person: Person, ctx: RequestContext): Person => {
	// do something
    }
}
```

The `read` and `write` rules are filter rules, while the `transform`, is accurately named. The transform rules will be implemented later.

The typescript in a filter rule is very limited. Only non-backward-jumping code is allowed, and currently, pretty much anything but `if-else` and `return` statement is not implemented.

When an apply request is received by the rpc server, all policies are collected, and the one containing typescript (ending in `.ts`), are passed to the policy compiler. The policy first generate a controlflow graph of the code, which it uses to check invariants (TODO). The controlflow graph is then split into regions, and the d_ir is constructed. For now the d_ir has no practical uses, but it'll come handy later for analysis.

From the controlflow graph, we obtain boolean expression for each `Action` (`Allow`, `Skip`, `Deny`, ..). From this boolean expression, we determine a SQL select clause for all the entities that need to filtered out (the inverse of the boolean expression for `Skip`).
The other use for the boolean expression, is to apply the restrictive read rules to the write rules (TODO).


chiselc/src/tools/d_ir.rs
```diff
@@ -0,0 +1,629 @@
...
+    PatOrExpr, Stmt, VarDecl,
+};
+
+use super::control_flow::Idx;
+use super::region::Region;
+use super::stmt_map::StmtMap;
+
+type Graph = StableDiGraph<Node, Edge>;
+
+#[derive(Default, Debug, Clone)]
```
penberg: Please add a Rustdoc here explaining what "ve map" is and what it is used for.

chiselc/src/tools/d_ir.rs
```diff
@@ -0,0 +1,629 @@
...
+pub(crate) struct VeMap {
+    map: HashMap<Symbol, NodeIndex>,
+}
+
```
penberg: A comment explaining what "blackbox" is would be helpful.

chiselc/src/tools/d_ir.rs
```diff
@@ -0,0 +1,629 @@
...
+#[allow(dead_code)]
+#[derive(Debug)]
+pub struct EnrichedRegion {
+    pub(crate) ve_map: VeMap,
+    context: GraphContext,
+    pub inner: Box<EnrichedRegionInner>,
+}
+
```
penberg: Please add a Rustdoc explaining what "DIr" is and what it is used for.

penberg (commented):
Perhaps bike-shedding, but `read`, `write`, and `transform` seem too low-level names to me. My suggestions would be `query`, `update`, and `map`. That said, no need to change anything right now if you prefer the names you already have.

Few comments on the commits.

MarinPostma:
@penberg yeah, while it remains in experimental phase we can rename anything, that's also what's cool with experimental features :)

honzasp:
Should we parse the policies on the server? I think it would make more sense to do the parsing and analysis in `chisel apply` (like we do with everything else).

Also, how are the entity policies preserved? I'm not sure I understand the code, but it looks like the entity policy is not stored in the database, so it won't be present when the server restarts?

MarinPostma:
@honza, this is not yet implemented, they are stored on the server

MarinPostma:
@honzasp as for policies being parsed on the cli, I kind of agree, but it is just more natural to follow the current policies flow and do it server side. The goal is to deprecate them, and then it'll probably be a good idea to move back this code to the cli.

glommer:
Just to align all reviewers, things like database persistence for this can come way later. We should be able to demonstrate some cool policies with this before we commit, so for now should focus on ergonomics, expressiveness, and ability to do database translations.

Great work, Marin.

penberg (commented):
From my perspective, this can go in. It's mostly new code and, therefore, low risk. However, it's a lot of code, so let's give some time for others to have a look at this.

MarinPostma:
this is not production ready in any shape or form, not this is the final design. The goal of this pr is to bring in support code, so I can carry on integration and figure out the best way to do it. This is bound to change a lot, but this can now be done in more manageable units of PR.

glommer (approved):

PR #1738, opened 2022-08-31
Upgrade deno to v1.25.0

penberg:


honzasp (approved):
Looks good. Btw, what is the reason to embed Deno in our repository instead of depending on the crates from crates.io?

penberg:
@honzasp @BearLemma I was briefly debugging this and it seems as if `property_expr_to_string()` never gets called. Any ideas how Deno upgrade would have anything to do with this?

penberg:
@BearLemma Historically it was because we carried some Deno patches that were not released. That's long gone now, but I think it's a submodule still because we embed files directly in https://github.com/chiselstrike/chiselstrike/blob/main/tsc_compile_build/src/lib.rs. No objections to getting rid of the dependency.

PR #1743, opened 2022-09-02
Deny auth endpoints when AUTH_SECRET is absent

dekimir:
In 5a98be7 we forbade auth-endpoint access without the ChiselAuth
header.  But a bug snuck in where a request WITH that header would
succeed even if CHISELD_AUTH_SECRET was absent.  This patch fixes it by
requiring the header value be strictly equal to the secret value.

A couple of policies_loggedin tests were relying on the buggy behavior
and needed fixing, too.

### DIFF
--- a/server/src/deno.rs
+++ b/server/src/deno.rs
@@ -1271,10 +1271,8 @@ async fn special_response(
             .get("CHISELD_AUTH_SECRET")
             .cloned();
         match (expected_secret, auth_header) {
-            (Some(serde_json::Value::String(s)), Some(h)) if s != *h => {
-                return Ok(Some(ApiService::forbidden("Fundamental auth")?))
-            }
-            _ => (),
+            (Some(serde_json::Value::String(s)), Some(h)) if s == *h => (),
+            _ => return Ok(Some(ApiService::forbidden("Fundamental auth")?)),
         }
     } else {
         let username = get_username_from_id(state.clone(), userid.clone()).await;
--- a/cli/tests/integration_tests/rust_tests/auth.rs
+++ b/cli/tests/integration_tests/rust_tests/auth.rs
@@ -2,6 +2,27 @@ use crate::framework::prelude::*;
 
 #[chisel_macros::test(modules = Node)]
 pub async fn next_auth_crud(mut c: TestContext) {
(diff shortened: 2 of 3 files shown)


server/src/deno.rs
```diff
@@ -1271,10 +1271,8 @@ async fn special_response(
             .get("CHISELD_AUTH_SECRET")
             .cloned();
         match (expected_secret, auth_header) {
```
honzasp, line 1273: I think that the `if auth_header.is_none()` check, which happens a few lines above, is now redundant?

(Github does not allow me to add the comment to that line, unfortunately.)

server/src/deno.rs
```diff
@@ -1271,10 +1271,8 @@ async fn special_response(
...
-            (Some(serde_json::Value::String(s)), Some(h)) if s != *h => {
-                return Ok(Some(ApiService::forbidden("Fundamental auth")?))
-            }
-            _ => (),
+            (Some(serde_json::Value::String(s)), Some(h)) if s == *h => (),
+            _ => return Ok(Some(ApiService::forbidden("Fundamental auth")?)),
```
honzasp, line 1275: Maybe we could change the error message? I have no idea what `"Fundamental auth"` means, so I guess that our users will be even more confused.

server/src/deno.rs
dekimir: It is, but I kept it because it exits early, avoiding the secrets check.

server/src/deno.rs
dekimir: I deliberately kept the messages curt as a defensive measure against attackers.  A message helpful to the legitimate user is also helpful to the attacker.  I thought users could ask us for translation in this case.

honzasp: Well, this code is open-source, so the "security by obscurity" argument does not seem to apply here. For the attacker, it will be trivial to grep the source code to find out what's going on, but legitimate users will be totally confused by this cryptic message.

glommer: agree with honza, but message is pre-existing, so we can and should change in a separate PR

glommer (approved):

server/src/deno.rs
dekimir: #1756

PR #1749, opened 2022-09-05
Rewrite filtering.deno to rust.

BearLemma:


cli/tests/integration_tests/framework.rs
```diff
@@ -442,7 +442,7 @@ impl Chisel {
     }
 
     pub async fn post_json_text<V: Borrow<serde_json::Value>>(&self, url: &str, data: V) -> String {
-        self.post(url).json(data).send().await.text()
+        self.post(url).json(data).send().await.assert_ok().text()
```
honzasp, line 445: I think that you should now rename `post_json_ok()` to `post_json()`, because the `_ok` becomes implicit.

cli/tests/integration_tests/rust_tests/api_filtering.rs
```diff
@@ -0,0 +1,271 @@
...
+                .toArray();
+        }"#,
+    );
+    c.chisel.apply_ok().await;
+    store_people(&c.chisel).await;
+
+    assert_eq!(
+        c.chisel.get_json("/dev/query?first_name=Glauber").await,
+        json!(["Costa"])
+    );
```
honzasp, line 84: Did you consider using `assert_json()` here? If so, what is the reason for you to prefer an explicit `assert_eq!`?

MarinPostma (approved):

cli/tests/integration_tests/rust_tests/api_filtering.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    assert_eq!(
+        c.chisel.get_json("/dev/query?human=false").await,
+        json!(["Pekka"])
+    );
+    let mut humans: Vec<String> =
+        serde_json::from_value(c.chisel.get_json("/dev/query?human=true").await).unwrap();
+    humans.sort();
+    assert_eq!(humans, vec!["Glauber", "Jan"]);
```
honzasp, line 159: Hmm, do you thing this is going to be more common? We could add a helper method for this (a natural generalization of the `.json()` method)

cli/tests/integration_tests/rust_tests/api_filtering.rs
```diff
@@ -0,0 +1,271 @@
...
+        r#"
+        import { Foo } from '../models/foo.ts';
+
+        export default async (req: Request) => {
+            await Foo.build({ f1: "x", f2: 1 }).save();
+            await Foo.build({ f1: "y", f2: 2 }).save();
+            await Foo.build({ f1: "z", f2: 3 }).save();
+            let c = await Foo.cursor().sortBy("f2");
+
+            const simple_take = await c
```
honzasp, line 189: Just a nitpick, TypeScript uses camelCase, so this should be `simpleTake`

cli/tests/integration_tests/rust_tests/api_filtering.rs
```diff
@@ -0,0 +1,271 @@
...
+                .take(1)
+                .toArray();
+            const take_filter = await c
+                .take(1)
+                .filter({ f1: "z" })
+                .toArray();
+            const filter_take = await c
+                .filter({ f1: "z" })
+                .take(1)
+                .toArray();
```
honzasp, line 199: In this test, the `take()` does nothing, because the cursor already contains just a single item. Perhaps we should add another entity with `{ f1: "z" }` to the database, to make sure that the `take(1)` really has an effect?

cli/tests/integration_tests/rust_tests/api_filtering.rs
```diff
@@ -0,0 +1,271 @@
...
+            await Foo.build({ f1: "z", f2: 3 }).save();
+            let c = await Foo.cursor().sortBy("f2");
+
+            const simple_take = await c
+                .take(1)
+                .toArray();
+            const take_filter = await c
+                .take(1)
+                .filter({ f1: "z" })
+                .toArray();
```
honzasp, line 195: This test is quite weak, because it would pass if:
- The `take()` skipped all items
- The `filter()` skipped all items

I think that it would be better if there were three items with `{ f1: "z" }` in the database:
- First would be skipped by the `take()`
- Third would be skipped by the `filter()`
- Second would be returned

cli/tests/integration_tests/rust_tests/policies_labels.rs
```diff
@@ -56,6 +56,16 @@ lazy_static::lazy_static! {
     });
 }
 
+lazy_static::lazy_static! {
```
honzasp, line 59: Do you think it would be a good idea to put the `lazy_static!` into `framework::prelude`, like `json!`?

cli/tests/integration_tests/rust_tests/policies_labels.rs
```diff
@@ -263,3 +273,56 @@ async fn persistence_after_restart(mut c: TestContext) {
...
+    assert!(r["results"].as_array().unwrap().len() == 1);
+    json_is_subset(
+        &r,
+        &json!({"results": [
+            {"first_name": "Pekka", "last_name": "xxxxx"}
+        ]}),
+    )
+    .unwrap();
+
+    // TODO: Is this correct? We are leaking the presence of Costa in the data.
```
honzasp, line 318: This is a very important point, perhaps we should escalate to Dejan? cc @dekimir

honzasp:
Uhm, the PR was merged even though there are unresolved review comments :(

honzasp:
Ah, apparently, I reviewed an already approved PR :)

cli/tests/integration_tests/rust_tests/policies_labels.rs
honzasp: This is a big issue, because in the presence of comparison operators, the attacker can binary-search for the value of the "anonymized" field.

cli/tests/integration_tests/framework.rs
honzasp: I agree, I just suggest that you rename `post_json_ok` (which POSTs a JSON and checks that the response is ok) to `post_json`, because _all_ helper functions now check that the response is ok, so the `_ok` suffix is redundant.

cli/tests/integration_tests/rust_tests/policies_labels.rs
MarinPostma: I actually think that the solution here is to check when a transform is applied to a field, and ignore that field in queries filters on that field altogether. If we decide to perform filters on the transformed data, then we end up with a db scan.

I think it's worth opening an issue on this matter, but this is of course outside the scope of this pr

dekimir: IIRC, when `anonymize` was first implemented, it wasn't possible to filter by an anonymized field.  I think this example shouldn't work.

PR #1750, opened 2022-09-05
Check that field type and it's default value have matching types.

### LINKED ISSUE #1598: Type annotation in model is ignored when there is a default value
In this model, the field `isCool` is inferred to have type `number` (from the default value), even though the explicit type annotation says `boolean`:

```ts
import { ChiselEntity } from '@chiselstrike/api';

export class Human extends ChiselEntity {
    name: string;
    isCool: boolean = 1;
}
```

Expected behavior is either that `isCool` is a `boolean` or an error. Note that typechecking fails in this case.

BearLemma:
Closes #1598.

### DIFF
--- a/cli/tests/integration_tests/rust_tests/model_defaults.rs
+++ b/cli/tests/integration_tests/rust_tests/model_defaults.rs
@@ -260,3 +260,81 @@ pub async fn complex_optional_defaults(c: TestContext) {
     )
     .unwrap();
 }
+
+#[chisel_macros::test(modules = Deno)]
+pub async fn field_type_and_value_type_mismatch(c: TestContext) {
+    c.chisel.write(
+        "models/default.ts",
+        r##"
+        import { ChiselEntity } from "@chiselstrike/api";
+
+        export class Defaulted extends ChiselEntity {
+            a: string = 1;
+        }"##,
+    );
+    c.chisel
+        .apply_err()
+        .await
+        .stderr
+        .read("field `a` is of type string but is default initialized by a value of type number");
+
+    c.chisel.write(
+        "models/default.ts",
+        r##"
+        import { ChiselEntity } from "@chiselstrike/api";
+
+        export class Defaulted extends ChiselEntity {
+            a: string = true;
+        }"##,
+    );
+    c.chisel
+        .apply_err()
+        .await
+        .stderr
+        .read("field `a` is of type string but is default initialized by a value of type boolean");
+
+    c.chisel.write(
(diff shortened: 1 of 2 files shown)


cli/tests/integration_tests/rust_tests/model_defaults.rs
```diff
@@ -260,3 +260,65 @@ pub async fn complex_optional_defaults(c: TestContext) {
...
+    );
+    c.chisel.apply_err().await;
+
+    c.chisel.write(
+        "models/default.ts",
+        r##"
+        import { ChiselEntity } from "@chiselstrike/api";
+
+        export class Defaulted extends ChiselEntity {
+            a: bool = "true";
```
honzasp, line 309: The correct syntax is `boolean`, so this file produces an error, but it is probably different from the error that we want to test for. Perhaps we should check the error returned from `apply_err().await`, to make the tests stronger?

cli/tests/integration_tests/rust_tests/model_defaults.rs
```diff
@@ -260,3 +260,65 @@ pub async fn complex_optional_defaults(c: TestContext) {
     )
     .unwrap();
 }
+
+#[chisel_macros::test(modules = Deno)]
+pub async fn field_type_and_value_type_mismatch(c: TestContext) {
```
honzasp, line 265: What happens when I use a more complex expression, for example `x: string = 1 + 1` or `x: number = function() {}`? Do we test this somewhere else? If not, perhaps we should add these tests here.

honzasp: Ok, what about the `Expr::Unary` case? If I look at the code, it seems that we would accept `x: number = !42` or `x: boolean = typeof true`?

honzasp: Are there some tests for that case?

honzasp (approved):

cli/tests/integration_tests/rust_tests/model_defaults.rs
dekimir: > Ok, what about the `Expr::Unary` case? If I look at the code, it seems that we would accept `x: number = !42` or `x: boolean = typeof true`?

#1291

dekimir: Ideally, the parser would recognize a negative number for what it is: a literal, not an expression. IDK if that's a viable option.

PR #1751, opened 2022-09-05
Improve take_and_filter_permutation test.

BearLemma:


cli/tests/integration_tests/rust_tests/api_filtering.rs
```diff
@@ -184,30 +184,31 @@ pub async fn take_and_filter_permutation(c: TestContext) {
...
+            await Foo.build({ f1: "z", f2: 4 }).save();
             let c = await Foo.cursor().sortBy("f2");
 
-            const simple_take = await c
+            const simpleTake = await c
                 .take(1)
                 .toArray();
-            const take_filter = await c
+            const takeFilter = await c
                 .take(1)
```
honzasp, line 194: Maybe it would be better to change this to `take(3)`, this will change the output to `[{f1: "z", f2: 3}]` and the test will become stronger.

honzasp: No, the output would have to be `[{f1: "z", f2: 3}]`:

- `c` is:
```
{f1: "x", f2: 1}
{f1: "y", f2: 2}
{f1: "z", f2: 3}
{f1: "z", f2: 4}
```

- `c.take(3)` is
```
{f1: "x", f2: 1}
{f1: "y", f2: 2}
{f1: "z", f2: 3}
```

- `c.filter({ f1: "z" })` is
```
{f1: "z", f2: 3}
```

If we didn't do SQL subqueries, the test would fail.

PR #1755, opened 2022-09-05
use separate .gen files for routes and events

### LINKED ISSUE #1754: chiselc mixing events and routes
Here's the result of looking in the sqlite database. There is a `hello.ts` in both events and routes in my filesystem:

```
> cat events/hello.ts
import { ChiselEvent } from "@chiselstrike/api";

function toJSON(buffer: ArrayBuffer) {
    return JSON.parse(String.fromCharCode.apply(null, new Uint8Array(buffer)));
}

export default async function (event: ChiselEvent) {
    console.log(toJSON(event.value));
}
```

```
cat routes/hello.ts
// Example ChiselStrike endpoint.
//
// To access the endpoint, run:
//
// curl -d '{"hello": "world"}' localhost:8080/dev/hello

export default async function (req: Request): Promise<string> {
    return await req.text() || "hello world";
}
```

Now looking at the `sources` table in sqlite:

```

select * from sources;
/dev/events/hello.ts|// .gen/hello.ts <== this is correct.
function toJSON(buffer) {
  return JSON.parse(String.fromCharCode.apply(null, new Uint8Array(buffer)));
}
async function hello_default(event) {
  console.log(toJSON(event.value));
}

// ../../var/folders/vy/4vtxx62s2cv2by55b_p29v300000gn/T/.tmpBRoUGh/3.ts
var __default = hello_default;
export {
  __default as default
};

/dev/routes/hello.ts|// .gen/hello.ts <=== see how this was overwritten by the events file
function toJSON(buffer) {
  return JSON.parse(String.fromCharCode.apply(null, new Uint8Array(buffer)));
}
async function hello_default(event) {
  console.log(toJSON(event.value));
}

// ../../var/folders/vy/4vtxx62s2cv2by55b_p29v300000gn/T/.tmpBRoUGh/1.ts
var __default = hello_default;
export {
  __default as default
};

```

glommer:
Fixes #1754

### DIFF
--- a/cli/src/cmd/apply/node.rs
+++ b/cli/src/cmd/apply/node.rs
@@ -43,11 +43,10 @@ pub(crate) async fn apply(
     }
 
     let cwd = env::current_dir()?;
-    let gen_dir = cwd.join(".gen");
-    fs::create_dir_all(&gen_dir)?;
 
     let mut chiselc_futures = vec![];
-    for endpoint in endpoints.iter().chain(events.iter()) {
+
+    let mut handle_code = |endpoint: &PathBuf, gen_dir: &PathBuf| {
         if optimize {
             let endpoint_file_path = endpoint.clone();
             let mut components = endpoint_file_path.components();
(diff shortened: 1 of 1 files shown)


dekimir (approved):

dekimir:
Follow-up: add a test.

PR #1756, opened 2022-09-06
More descriptinve error messages for ChiselAuth

dekimir:
https://github.com/chiselstrike/chiselstrike/pull/1743#discussion_r961494688

Make error messages more user-friendly when ChiselAuth is absent or
incorrect.

No new testcases; there already are tests for this, but they only check
the response status, to avoid fragility when messages evolve.


honzasp (approved):
Great! I think that this will improve the user experience :)

PR #1757, opened 2022-09-06
Fix convert_to_argument() macro argument type

penberg:
The "fallback" argument is a type, not an identifier. Noticed the latent
bug while attempting to pass a Vec<u8>.


MarinPostma:
@penberg what's the status on this one? it fails to build

penberg:
@MarinPostma Fixed now, thanks for the reminder!

MarinPostma (approved):

Issue #1759, opened 2022-09-07
Production Best Practices


7flash:
What's the recommended way to run chisel in production? Need to make sure exceptions are handled properly and chisel server can be auto-restarted when it fails. It happens sometimes it shows following message in logs and then stops running.

![image](https://user-images.githubusercontent.com/4569866/188820201-00aaad9a-24b0-43ca-9840-aaf12cd0df16.png)

dekimir:
@7flash please paste text instead of screenshots, it's much more useful.  This particular pic is cropped, missing a part of that warning.  I believe these warnings aren't bringing `chiseld` down, but I'll double-check after you paste the full text.

On our platform, we run `chiseld` in k8s pods and monitor it via these internal endpoints:
https://github.com/chiselstrike/chiselstrike/blob/a5aeb9970f38627f5939f761f0b86e961dc2d520/server/src/internal.rs#L29-L35

7flash:
Today it stopped with different error in logs:

![image](https://user-images.githubusercontent.com/4569866/189288423-1e0af3e8-ae95-4888-9daf-8f065f227702.png)

Running with pm2, but it didn't auto-restart

7flash:
The origin of the error above can be addressed in separate issue https://github.com/chiselstrike/chiselstrike/issues/1607

The question here, - is how to deal with general unexpected errors

pm2 auto-restart strategy expected to work with flag "--max-memory-restart 300M".. but not sure about that.. because chisel does instantiate a new process which might track memory separately and might not propagate error status to main process..

7flash:
Found solution to pm2 not restarting the process.. should be passing "--no-automation" flag

https://github.com/Unitech/pm2/issues/1934#issuecomment-184173656

7flash:
But then it fails with following error:

```
09-10 07:49: [2022-09-10T07:49:39Z] INFO - RPC is ready. URL: 127.0.0.1:10093
09-10 07:49: thread 'tokio-runtime-worker' panicked at 'error binding to 127.0.0.1:10091: error creating server listener: Address already in use (os error 98)', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/hyper-0.14.20/src/server/server.rs:77:13
09-10 07:49: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[...]

PR #1768, opened 2022-09-11
pm2 compatible #1759

7flash:


penberg:
@7flash Looks reasonable and am happy to merge this, but can you point me to some material on why this is needed? (I'm new to `pm2`...)

7flash:
https://github.com/Unitech/pm2/blob/da59cb6dd761546686e5f89dbc8126672d8b3460/types/index.d.ts#L381

pm2 is waiting for event from the process to consider it's active, otherwise it's thinking it has failed and trying to restart it, that's what happens in #1759 issue

normally, http module will emit "listen" event, but since we are spawing a separate process, we need to emit "wait" event manually and run pm2 with flag "wait_ready"

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Tags:** Inefficient

packages/chiselstrike-cli/run.js
```diff
@@ -2,3 +2,5 @@
 
 const { run } = require("./binary");
 run();
+
+process.send("ready");
```
penberg, line 6: I tested this locally and noticed that with the regular `npm run dev` flow (that `create-chiselstrike-app` generates, for example) never actually executes this line (`run()` call above never returns). Are you sure the call is in the right place?

Also, with `npm run dev` there is no IPC channel, which means `process` is `undefined` (https://nodejs.org/api/process.html#processsendmessage-sendhandle-options-callback) and that call would anyway fail so I think we need to first ensure `process` is actually set before calling `send`, no?

penberg (changes requested):

packages/chiselstrike-cli/run.js
glommer: @7flash any comments on this ?

7flash: that's right, fixed

glommer: Thanks. The linter complained, though. I think due to spaces.

You can run the linter locally to see what's up

PR #1770, opened 2022-09-12
add nocapture to test framework

MarinPostma:
adds a `nocapture` flag to the test framework. This flag is compliant with rust test interface, and can be passed like so:

```
cargo test -- --nocapture
```

When it is raised, all output from chiseld is echoed to stdout.

### DIFF
--- a/cli/tests/integration_tests/framework.rs
+++ b/cli/tests/integration_tests/framework.rs
@@ -1,16 +1,20 @@
 use crate::database::Database;
 use anyhow::{anyhow, Context, Result};
-use bytes::{Bytes, BytesMut};
+use bytes::{Buf, Bytes, BytesMut};
+use futures::future::poll_fn;
+use futures::{pin_mut, ready, Future, FutureExt};
 use std::borrow::Borrow;
+use std::io::{stdout, Write};
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::process::{ExitStatus, Stdio};
 use std::sync::Arc;
+use std::task::Poll;
 use std::time::Duration;
-use std::{error, fmt, str};
+use std::{error, fmt, io, str};
 use std::{fs, net::SocketAddr};
 use tempdir::TempDir;
-use tokio::io::{AsyncRead, AsyncReadExt, AsyncWriteExt};
+use tokio::io::{duplex, AsyncRead, AsyncReadExt, AsyncWriteExt, DuplexStream};
 
 pub mod prelude {
     pub use super::{json_is_subset, Chisel, Response, TestContext};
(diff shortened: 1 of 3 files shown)


cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +24,151 @@ pub mod prelude {
...
+                for _ in 0..32 {
+                    let reader = &mut self.reader;
+                    pin_mut!(reader);
+                    let read = reader.read_buf(&mut *buffer);
+                    pin_mut!(read);
+                    match read.poll(cx) {
+                        Poll::Ready(Ok(0)) => return Poll::Ready(Ok(())),
+                        Poll::Ready(Ok(n)) => {
+                            let len = buffer.len();
+                            let s = std::str::from_utf8(&buffer[len - n..]).unwrap();
```
honzasp, line 63: We should not panic when `buffer[len - n..]` is not valid utf-8. Even if the output of the program was utf-8 (which is not guaranteed!), the slice does not have to be valid utf-8, because it might slice a multi-byte character in half.

cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +24,151 @@ pub mod prelude {
...
 
+struct Tee {
+    buffer: Arc<Mutex<BytesMut>>,
+    reader_handle: tokio::task::JoinHandle<io::Result<()>>,
+}
+
+struct TeeReader<R> {
+    reader: R,
+    buffer: Arc<Mutex<BytesMut>>,
+}
```
honzasp, line 35: I don't understand this mechanism. I would think that it would simply be enough to wrap the reader `R` and just print all bytes read from `R` to stdout/stderr? Why is it necessary to spawn a task and store the data in the buffer?

cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +24,151 @@ pub mod prelude {
...
+        pin_mut!(fut);
+        let polled = fut.poll(cx);
+        match polled {
+            Poll::Ready(mut buffer) => {
+                for _ in 0..32 {
+                    let reader = &mut self.reader;
+                    pin_mut!(reader);
+                    let read = reader.read_buf(&mut *buffer);
+                    pin_mut!(read);
+                    match read.poll(cx) {
```
honzasp, line 59: I would consider using the macro `ready!` from the `futures` crate to streamline some of these matches (not only here but in other places in the code).

honzasp (commented):
This is very useful, I also found myself needing a mechanism like this, but I was too lazy to implement it properly, so thank you for tackling this! :) But it seems to me that the solution could be simplified (but I might be mistaken!)

cli/tests/integration_tests/framework.rs
MarinPostma: yeah right I should just pipe to stdout

cli/tests/integration_tests/framework.rs
MarinPostma: The goal here is to get as close as possible to real-time output. Without a polling task, someone would need to poll it, but who's duty would that be? The consumer of output? that would be strange wouldn't it? since we'd get nothing printed if the output is not checked. It seemed to me like a reasonable solution to drive this output.

honzasp: Ah, _this_ is the crucial piece of information that I did not understand :)

In this case, maybe it would be better to add this functionality to the `AsyncTestableOutput`? It already has a buffer, so we could add the `TaskHandle` to the `AsyncTestableOutput` and spawn the reading task in its `new()` function? This might simplify the code, because we would not have to introduce two structs and implement `AsyncRead`.

MarinPostma: I'm not convinced, the logic of AsyncRead would need to be somewhere anyway. At least here, this is some reusable and isolated code whose purpose is clear.

cli/tests/integration_tests/main.rs
```diff
@@ -71,6 +71,10 @@ pub struct Opt {
     #[structopt(short, long)]
     pub parallel: Option<usize>,
     test_arg: Option<Regex>,
+
+    /// don't capture stdout/stderr of each task, allow printing directly
+    #[structopt(long)]
+    nocapture: bool,
```
honzasp, line 77: The naming is slightly confusing; we _always_ capture the stdout/stderr of `chiseld`, what this flag controls is whether we also duplicate the data the the test suite's stdout/stderr. What about something like `--show-stdout`, `--debug-chiseld` or `--show-chiseld`?

MarinPostma: This is to be compliant with rust arguments. Everytime we decide to add an argument that is not natively supported by rust tests, we make it harder to run tests for ourselves

cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +25,146 @@ pub mod prelude {
...
 }
 
+struct Tee {
+    buffer: Arc<Mutex<BytesMut>>,
+    reader_handle: tokio::task::JoinHandle<io::Result<()>>,
+}
+
+struct TeeReader<R> {
+    reader: R,
+    buffer: Arc<Mutex<BytesMut>>,
```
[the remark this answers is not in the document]
MarinPostma: I am never holding the lock across await

cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +25,146 @@ pub mod prelude {
...
+                    pin_mut!(handle);
+                    match handle.poll(cx) {
+                        Poll::Ready(Ok(Ok(()))) => Poll::Ready(Ok(())),
+                        Poll::Ready(Ok(Err(e))) => Poll::Ready(Err(e)),
+                        Poll::Ready(Err(e)) => {
+                            Poll::Ready(Err(io::Error::new(io::ErrorKind::Other, e)))
+                        }
+                        Poll::Pending => Poll::Pending,
+                    }
+                }
```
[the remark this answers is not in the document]
MarinPostma: I have found a better way

cli/tests/integration_tests/main.rs
```diff
@@ -71,6 +71,10 @@ pub struct Opt {
     #[structopt(short, long)]
     pub parallel: Option<usize>,
     test_arg: Option<Regex>,
+
+    /// When raised, echoes chiseld output to stdout.
```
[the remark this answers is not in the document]
MarinPostma: I have updated it to be the exact same doc as the one from rust tests

cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +25,115 @@ pub mod prelude {
...
+        let (input, mut output) = duplex(16384);
+
+        let reader_handle = tokio::spawn(async move {
+            let mut buffer = BytesMut::new();
+            loop {
+                match reader.read_buf(&mut buffer).await {
+                    Ok(0) => break Ok(()),
+                    Ok(count) => {
+                        let new_bytes = &buffer[buffer.len() - count..];
+                        stdout().write_all(new_bytes).unwrap();
```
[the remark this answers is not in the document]
MarinPostma: Stdout usually doesn't block, i mean we also use it for logging, and nobody bats an eye right?

cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +25,115 @@ pub mod prelude {
...
+    {
+        let (input, mut output) = duplex(16384);
+
+        let reader_handle = tokio::spawn(async move {
+            let mut buffer = BytesMut::new();
+            loop {
+                match reader.read_buf(&mut buffer).await {
+                    Ok(0) => break Ok(()),
+                    Ok(count) => {
+                        let new_bytes = &buffer[buffer.len() - count..];
```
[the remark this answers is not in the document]
MarinPostma: It's because this is a buf writer, think of it like a ring buffer

cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +25,115 @@ pub mod prelude {
...
+        R: AsyncRead + Unpin + Send + 'static,
+    {
+        let (input, mut output) = duplex(16384);
+
+        let reader_handle = tokio::spawn(async move {
+            let mut buffer = BytesMut::new();
+            loop {
+                match reader.read_buf(&mut buffer).await {
+                    Ok(0) => break Ok(()),
+                    Ok(count) => {
```
honzasp, line 45: Just a small suggestion, we could reduce the nesting by saying `while let count = reader.read_buf().await? { if count == 0 { break } ... }` (but this is just a stylistic choice, feel free to ignore :) )

cli/tests/integration_tests/framework.rs
```diff
@@ -21,28 +25,115 @@ pub mod prelude {
...
+                                let fut = output.write_buf(&mut buffer);
+                                pin_mut!(fut);
+                                match fut.poll(cx) {
+                                    Poll::Ready(_) => continue,
+                                    Poll::Pending => return Poll::Ready(()),
+                                }
+                            }
+                            Poll::Ready(())
+                        })
+                        .await;
```
honzasp, line 60: What is the reason to use `poll_fn` instead of simply `output.write_all_buf(&mut buffer).await?`? I admit that I was not able to understand the comment, though.

cli/tests/integration_tests/main.rs
honzasp: Hmm, how does adding an extra argument make tests harder to run? I always run the tests using `cargo -p cli --test integration_tests --`, where I can pass any arguments; do you use something else? Also, the `--nocapture` flag in Rust does something different, it applies to the code in the test function, while this flag applies only to the output of `chiseld`.

However, this is not so important, so if you still disagree, keep it as it is :)

MarinPostma:
@honzasp very simple, my claim is that this syntax:
`cargo test integration::some::test -- --nocapture`
is simple than:
`cargo test -p cli --test integration_tests -- --test some::test --nocapture`

if you decide to change `--nocapture` to some non-conventional argument, then the first test bin will reject it (most likely unit tests), and now you have to type this crazy long command that I keep forgetting.

It might not be _exactly_ what `nocapture` does in rust, but it's similar enough that you'd think to use it in this case.

cli/tests/integration_tests/framework.rs
honzasp: Hmm, but this means that if the duplex becomes full and the consumer then reads the bytes, then we won't fill it again until we read some more bytes from `reader`; maybe we should use `select!` to concurrently read from `reader` and write to `output`?

honzasp:
Hmm, so when you want to run a particular integration test, you run _all_ tests and the other tests will just skip all their test cases?

PR #1772, opened 2022-09-13
Use ChiselRequest in hello.ts template.

BearLemma:
It's better to instruct users to use `ChiselRequest` which is more awesome.

### DIFF
--- a/packages/create-chiselstrike-app/template/hello.ts
+++ b/packages/create-chiselstrike-app/template/hello.ts
@@ -3,7 +3,8 @@
 // To access the endpoint, run:
 //
 // curl -d '{"hello": "world"}' localhost:8080/dev/hello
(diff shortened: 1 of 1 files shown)


glommer:
why? @CodingDoug and I just had this same discussion somewhere else, and it's best to optimize for simplicity, not awesomeness =)

CodingDoug:
I hadn't looked at ChiselRequest until now, and now that I see that it does add some information about the request, I think it's a good idea to make this the recommended type for endpoints/routes throughout the docs.  For the tutorial, I will take Jan's suggestion and only state that it's a subclass of Request.  Elsewhere in the reference documentation I will explain in more detail what ChiselRequest offers.

glommer (approved):

PR #1773, opened 2022-09-13
fix framework nocapture

MarinPostma:
As @honzasp noticed, there was an issue in the implementation on `Tee`, where we'd stop writing to the `DuplexStream` if it was full, and not write to it again until we could read again from the reader.

@honzasp suggested that we `select`, but that would mean that we'd need to borrow the buffer mutably twice.

This leaves us with the option of implementing the future ourselves.

### DIFF
--- a/cli/tests/integration_tests/framework.rs
+++ b/cli/tests/integration_tests/framework.rs
@@ -1,12 +1,9 @@
 // SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
 
-use crate::database::Database;
-use anyhow::{anyhow, Context, Result};
-use bytes::{Buf, Bytes, BytesMut};
-use futures::future::poll_fn;
-use futures::{pin_mut, ready, Future, FutureExt};
 use std::borrow::Borrow;
+use std::fs;
 use std::io::{stdout, Write};
+use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 use std::pin::Pin;
 use std::process::{ExitStatus, Stdio};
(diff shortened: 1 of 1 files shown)


cli/tests/integration_tests/framework.rs
```diff
@@ -28,43 +31,82 @@ pub mod prelude {
...
-    reader_handle: tokio::task::JoinHandle<io::Result<()>>,
+    reader_handle: Fuse<tokio::task::JoinHandle<io::Result<()>>>,
     input: DuplexStream,
 }
 
+struct TeeReader<R> {
+    output: DuplexStream,
+    buffer: BytesMut,
+    reader: R,
+    should_return: bool,
```
honzasp, line 42: I suggest renaming this to `eof` to make it more clear what this flag signifies.

cli/tests/integration_tests/framework.rs
```diff
@@ -28,43 +31,82 @@ pub mod prelude {
...
+        }
+
+        if buffer.has_remaining() {
+            let fut = output.write_all_buf(buffer);
+            pin_mut!(fut);
+            if let Poll::Ready(res) = fut.poll(cx) {
+                match res {
+                    // If we succeed here, we must poll ourselves immediately to poll the reader
+                    // again.
+                    Ok(_) => cx.waker().wake_by_ref(),
```
honzasp, line 82: Why don't we use an explicit loop? This would be cleaner and slightly more efficient.

MarinPostma: uhm, yes, but the loop would have to break when the two return pending? I can try

honzasp (approved):

PR #1774, opened 2022-09-13
Add copyright notices where missing

dekimir:
Many files were missing a copyright notice.


penberg:
@BearLemma @dekimir I think we should revert this change. It adds our copyright to TSC test suite files...

penberg:
Never mind, I was confused. The tests are in `tsc_compile` which is *not*  `tsc`, but our own wrapper. Sorry for the noise!

PR #1775, opened 2022-09-14
Reject unexpected YAML elements in policy files

dekimir:
It's too easy to mistype some YAML and fail to apply a policy, without any warning or error.

### DIFF
--- a/server/src/policies.rs
+++ b/server/src/policies.rs
@@ -225,6 +225,10 @@ impl VersionPolicy {
 
     fn add_labels(&mut self, labels: &[Yaml]) -> Result<()> {
         for label in labels.iter() {
+            if label.as_hash().is_none() {
+                anyhow::bail!("label not a dictionary: {label:?}");
+            }
+
             let name = label["name"].as_str().ok_or_else(|| {
                 anyhow::anyhow!("couldn't parse yaml: label without a name: {:?}", label)
             })?;
--- a/server/src/policies.rs
+++ b/server/src/policies.rs
@@ -229,9 +229,11 @@ impl VersionPolicy {
                 anyhow::bail!("label not a dictionary: {label:?}");
             }
 
-            let name = label["name"].as_str().ok_or_else(|| {
-                anyhow::anyhow!("couldn't parse yaml: label without a name: {:?}", label)
-            })?;
+            let name = match &label["name"] {
+                Yaml::String(s) => s,
+                Yaml::BadValue => anyhow::bail!("label without a name: {label:?}"),
+                x => anyhow::bail!("label name isn't a string: {x:?}"),
+            };
 
             debug!("Applying policy for label {:?}", name);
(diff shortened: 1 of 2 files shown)


MarinPostma:
@dekimir, it seems to me that the validation that you are performing could be performed by serde, what exactly prevents us to do that?

MarinPostma:
Alright, I'll take a deeper look at this pr in the mean time

honzasp:
I think that `serde` ignores unknown keys by default, but `#[serde(deny_unknown_fields)]` should fix it. (Though I'm not sure whether the error messages will be clear to the user!)

server/src/policies.rs
```diff
@@ -199,75 +199,116 @@ impl Policies {
...
+                        let skey = key.as_str().unwrap_or("");
+                        match (skey, value) {
+                            ("labels", Yaml::Array(labels)) => policies.add_labels(labels)?,
+                            ("routes" | "endpoints", Yaml::Array(routes)) => {
+                                policies.add_routes(routes)?
+                            }
+                            ("labels" | "routes" | "endpoints", _) => {
+                                bail!("incorrect value for {skey}: {value:?}")
+                            }
+                            _ => bail!("unexpected yaml key: {key:?}"),
```
MarinPostma, line 216: just a nit, but I not a fan of unwrapping to an invalid value before matching on it. What do you think of that:
```suggestion
                        let skey = key.as_str();
                        match (skey, value) {
                            (Some("labels"), Yaml::Array(labels)) => policies.add_labels(labels)?,
                            (Some("routes" | "endpoints"), Yaml::Array(routes)) => {
                                policies.add_routes(routes)?
                            }
                            (Some("labels" | "routes" | "endpoints")), _) => {
                                bail!("incorrect value for {key}: {value:?}")
                            }
                            _ => bail!("unexpected yaml key: {key:?}"),
```

server/src/policies.rs
```diff
@@ -199,75 +199,116 @@ impl Policies {
...
+                    }
+                }
+                _ => bail!("top-level policies yaml isn't a dictionary: {config:?}"),
+            }
+        }
+        Ok(policies)
+    }
+
+    fn add_labels(&mut self, labels: &[Yaml]) -> Result<()> {
+        for label in labels.iter() {
```
MarinPostma, line 227:
```suggestion
        for label in labels {
```

honzasp:
> And got error:
> 
> ```
> Error("unknown field `trol`, expected one of `name`, `age`, `phones`", line: 5, column: 18)
> ```
> 
> which I think is pretty good.

Very nice! Does it work the same with yaml?

MarinPostma:
@honzasp i would suppose so, this error is generated by the derives implementation of deserialize by serde, not serde_yaml

dekimir:
Replaced by #1809, which copies all the parsing tests from here but uses serde_yaml to parse.

PR #1776, opened 2022-09-19
Replace npx with node in README

psqli:
npx does not work with paths

In Node 14.x CLI documentation it's unclear that it's valid to pass a path to a package, however I've tested it and it seems to work as documented by Node 16.x documentation:  

	https://nodejs.org/docs/latest-v16.x/api/cli.html#command-line-api

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,9 +203,9 @@ cargo build
 
 That will build the `chiseld` server and `chisel` utility.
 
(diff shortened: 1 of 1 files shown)


psqli:

penberg (approved):
Looks good, thanks @psqli!

PR #1780, opened 2022-09-20
Entity value

BearLemma:
Introduces `EntityValue` paving way for #1785 . `EntityValue` is used to provide natural transition layer between JavaScript and Rust (and back).

### DIFF
--- a/api/src/datastore.ts
+++ b/api/src/datastore.ts
@@ -53,8 +53,9 @@ abstract class Operator<Input, Output> {
                 const rid = await getRid();
                 try {
                     while (true) {
-                        const properties = await opAsync(
-                            "op_chisel_query_next",
+                        await opAsync("op_chisel_query_next", rid);
+                        const properties = opSync(
+                            "op_chisel_query_get_value",
                             rid,
                         );
 
--- a/server/src/datastore/crud.rs
+++ b/server/src/datastore/crud.rs
@@ -3,11 +3,13 @@
 use crate::datastore::engine::{QueryEngine, TransactionStatic};
 use crate::datastore::expr::{BinaryExpr, BinaryOp, Expr, PropertyAccess, Value as ExprValue};
 use crate::datastore::query::{Mutation, QueryOp, QueryPlan, RequestContext, SortBy, SortKey};
+use crate::datastore::value::EntityMap;
 use crate::types::{Entity, Type, TypeSystem};
 use crate::JsonObject;
 use anyhow::{Context, Result};
 use deno_core::futures;
 use futures::{Future, StreamExt};
+use guard::guard;
 use serde_derive::{Deserialize, Serialize};
 use serde_json::json;
 use url::Url;
@@ -52,7 +54,7 @@ fn run_query_impl(
             .collect::<Vec<_>>()
             .await
             .into_iter()
-            .collect::<Result<Vec<_>>>()
+            .collect::<Result<Vec<EntityMap>>>()
             .context("failed to collect result rows from the database")?;
 
         // When backwards cursor is specified, the sort is reversed, hence we
@@ -62,9 +64,16 @@ fn run_query_impl(
                 results.reverse();
             }
         }
+        let results: Vec<_> = results
+            .iter()
+            .map(|entity_fields: &EntityMap| {
+                let v = serde_json::to_value(entity_fields).unwrap();
+                guard! {let serde_json::Value::Object(map) = v else { panic!("expected json object") }}
+                map
+            })
+            .collect();
 
         let mut ret = JsonObject::new();
-
         let next_page = get_next_page(&params, &query, &host, &results)?;
         if let Some(next_page) = next_page {
             ret.insert("next_page".into(), json!(next_page));
--- a/server/src/datastore/value.rs
+++ b/server/src/datastore/value.rs
@@ -1,0 +1,280 @@
+use anyhow::{bail, Context as _, Result};
+use deno_core::v8;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(untagged)]
+pub enum EntityValue {
+    Null,
+    String(String),
+    Float64(f64),
+    Boolean(bool),
+    // Representation of JavaScript's Date represented as UNIX timestamp,
+    // specifically it's the number of milliseconds since epoch.
+    JsDate(f64),
+    Array(EntityArray),
+    Map(EntityMap),
+}
+
+pub type EntityArray = Vec<EntityValue>;
+pub type EntityMap = HashMap<String, EntityValue>;
+
+impl EntityValue {
+    pub fn from_v8<'a>(
+        v: &v8::Local<'a, v8::Value>,
+        scope: &mut v8::HandleScope<'a>,
+    ) -> Result<EntityValue> {
+        let value = if v.is_null() {
+            EntityValue::Null
+        } else if v.is_string() {
+            EntityValue::String(v.to_string(scope).unwrap().to_rust_string_lossy(scope))
+        } else if v.is_number() {
+            EntityValue::Float64(v.to_number(scope).unwrap().value())
+        } else if v.is_boolean() {
+            EntityValue::Boolean(v.boolean_value(scope))
+        } else if v.is_date() {
+            let date = v8::Local::<v8::Date>::try_from(*v).unwrap();
+            EntityValue::JsDate(date.value_of())
+        } else if v.is_array() {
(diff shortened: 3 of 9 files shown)


server/src/datastore/value.rs
```diff
@@ -0,0 +1,357 @@
...
+        } else if v.is_boolean() {
+            EntityValue::Boolean(v.boolean_value(scope))
+        } else if v.is_date() {
+            let date = v8::Local::<v8::Date>::try_from(*v).unwrap();
+            let datetime = NaiveDateTime::from_timestamp(0, 0)
+                + Duration::milliseconds(date.value_of() as i64);
+            EntityValue::JsDate(datetime)
+        } else if v.is_array() {
+            let array = v8::Local::<v8::Array>::try_from(*v).unwrap();
+            let mut value_vec = vec![];
```
honzasp, line 100: As a tiny performance optimization, you might consider `let mut value_vec = Vec::with_capacity(array.length())` here.

server/src/datastore/value.rs
```diff
@@ -0,0 +1,357 @@
...
+            for i in 0..array.length() {
+                let val = array.get_index(scope, i).unwrap();
+                value_vec.push(Self::from_v8(&val, scope)?);
+            }
+            EntityValue::Array(value_vec)
+        } else if v.is_object() {
+            let obj = v.to_object(scope).unwrap();
+            let prop_names = obj.get_own_property_names(scope).unwrap();
+            let mut record = EntityMap::new();
```
honzasp, line 109: Same as above, `EntityMap::with_capacity(prop_names.length())` would be a tiny bit more performant.

server/src/datastore/value.rs
```diff
@@ -0,0 +1,357 @@
...
+    pub fn as_map(&self) -> Result<&EntityMap> {
+        if let Self::Map(v) = self {
+            Ok(v)
+        } else {
+            bail!(
+                "tried to convert entity value to EntityMap, but it is of type {}",
+                self.kind_str()
+            )
+        }
+    }
```
honzasp, line 273: Do you think that these methods could also be replaced by macros?

server/src/datastore/value.rs
```diff
@@ -0,0 +1,357 @@
...
+
+impl PartialEq<EntityValue> for String {
+    fn eq(&self, other: &EntityValue) -> bool {
+        eq_str(other, self.as_str())
+    }
+}
+
+macro_rules! partialeq_numeric {
+    ($($eq:ident [$($ty:ty)*])*) => {
+        $($(
```
honzasp, line 326: Instead of a single call to a variadic macro, it might be simpler to invoke a non-variadic macro multiple times, something like:
```
partialeq_numeric!(f32, eq_f64);
partialeq_numeric!(f64, eq_f64);
partialeq_numeric(bool, eq_bool);
```

server/src/ops/datastore.rs
```diff
@@ -293,19 +310,33 @@ impl Future for QueryNextFuture {
...
+) -> Result<serde_v8::Value<'a>> {
+    let rc: Rc<QueryStreamResource> = state.borrow().resource_table.get(query_stream_rid)?;
+    let value = match rc.next.borrow_mut().take() {
+        Some(v) => v.to_v8(scope)?.into(),
+        None => {
+            let null = v8::null(scope);
+            serde_v8::from_v8(scope, null.into())?
+        }
+    };
+    Ok(value)
```
honzasp, line 339: I think it should be possible to simplify this to:
```
let v8_value = match query_stream.next.borrow_mut().take() {
    Some(v) => v.to_v8(scope)?,
    None => v8::null(scope),
};
Ok(serde_v8::Value::from(v8_value))
```

api/src/datastore.ts
```diff
@@ -53,8 +53,9 @@ abstract class Operator<Input, Output> {
                 const rid = await getRid();
                 try {
                     while (true) {
-                        const properties = await opAsync(
-                            "op_chisel_query_next",
+                        await opAsync("op_chisel_query_next", rid);
+                        const properties = opSync(
+                            "op_chisel_query_get_value",
```
MarinPostma, line 58: I think this is incorrect. I can perfectly write the following code:
```typescript
let iter = runChiselQuery();
let items = Array.from(iter);
let results = Promise.all(items);
```

Since all the promises share the same rid to get their value, it is easy to see that this will lead to a data race.

server/src/datastore/crud.rs
```diff
@@ -62,9 +62,16 @@ fn run_query_impl(
                 results.reverse();
             }
         }
+        let results: Vec<_> = results
+            .iter()
+            .map(|v| {
+                let v = serde_json::to_value(v).unwrap();
+                // TODO: this is an unnecessary copy
+                v.as_object().unwrap().to_owned()
```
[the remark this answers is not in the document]
honzasp: Hmm, the previous line says `let v = serde_json::to_value(v).unwrap();`, shouldn't the `v` in `v.as_object()` be a `serde_json::Value`? (The `v` given to the closure `|v|` is an `EntityValue`, you might consider using two distinct names to make this more clear :) Some type annotation might also be helpful.)

server/src/datastore/value.rs
```diff
@@ -0,0 +1,357 @@
...
+
+#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+#[serde(untagged)]
+pub enum EntityValue {
+    Null,
+    String(String),
+    Float64(f64),
+    Boolean(bool),
+    #[serde(serialize_with = "to_timestamp")]
+    JsDate(NaiveDateTime),
```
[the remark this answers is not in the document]
honzasp: By accuracy I mean accurately adhering to the internal representation of JS `Date`, which is an f64. Do you think that `JsDate(f64)` is unclear? A possible solution would be to introduce `struct JsDate { timestamp_ms: f64 }`.

I don't have anything against `chrono`, but it seems unnecessary, given that we only use `NaiveDateTime` as a more readable and less efficient way to store a float. Of course, it would be different if we actually did some computations with the date times, but this does not seem to be the case.

server/src/datastore/value.rs
honzasp: I thought so :D We should be careful with this, though, because this is already some non-trivial snippet, so licensing and copyrights might apply :)

api/src/datastore.ts
honzasp: Hmm, does it? :thinking: It seems to me that we will always alternate the calls to `op_chisel_query_next`/`op_chisel_query_get_value` correctly. Note that the return value from `runChiselQuery()` is not an iterator of promises but an async iterator, so the body of the `while` loop should be executed sequentially.

server/src/datastore/value.rs
honzasp:
```
macro_rules! as_copy {
    ($method_name:ident, $variant:ident, $typ:ty) => {
        pub fn $method_name(&self) -> Result<$typ> {
            match self {
                Self::$variant(v) => Ok(*v),
                _ => bail!(
                    "tried to convert entity value to {}, but it is of type {}",
                    stringify!($typ),
                    self.kind_str(),
                ),
            }
        }
    }
}

macro_rules! as_ref {
    ($method_name:ident, $variant:ident, $typ:ty) => {
        pub fn $method_name(&self) -> Result<&$typ> {
            match self {
                Self::$variant(v) => Ok(v),
                _ => bail!(
                    "tried to convert entity value to {}, but it is of type {}",
                    stringify!($typ),
                    self.kind_str(),
                ),
            }
        }
    }
}

impl EntityValue {
    as_ref!(as_str, String, str);
    as_copy!(as_f64, Float64, f64);
    as_copy!(as_bool, Boolean, bool);
    as_copy!(as_js_date, JsDate, NaiveDateTime);
    as_ref!(as_array, Array, EntityArray);
    as_ref!(as_map, Map, EntityMap);
}
```
should do the trick.

server/src/ops/mod.rs
```diff
@@ -33,6 +33,7 @@ pub fn extension() -> deno_core::Extension {
             env::op_env::decl(),
             env::op_get_env::decl(),
             env::op_delete_env::decl(),
+            datastore::op_chisel_query_get_value::decl(),
```
honzasp, line 36: Hmm, this list was meant to be sorted, or at least grouped by the module :)

honzasp (approved):

api/src/datastore.ts
MarinPostma: well from js own definition, an async iterator is an iterator of promises: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_async_iterator_and_async_iterable_protocols. But for some wicked reason this seems to work 🤷

Issue #1530, opened 2022-07-07
Dynamic require of "buffer" is not supported


honzasp:
When I add the `memfs` package to `package.json` and import it (`import ... from memfs`), I get the following error:

```
> bomb@1.0.0 dev
> chisel dev

[2022-07-07T15:20:12Z] INFO - ChiselStrike is ready 🚀 - URL: http://localhost:8080 
Unhandled exception: Error: Promise rejected without a handler: Error: Dynamic require of "buffer" is not supported
    at file:///worker.js:6:15
parsing endpoints

Caused by:
    Error: Dynamic require of "buffer" is not supported
```

Steps to reproduce:
- Run `npx -y create-chiselstrike-app bomb`
- Add `"memfs": "^3.4.7"` to `"dependencies"` in `package.json`
- Run `npm install`
- Replace `endpoints/hello.ts` with the following
```typescript
import memfs from 'memfs';

export default async function (req: Request): Promise<Response> {
    return new Response(memfs.Volume);
}
```
- Start the backend with `npm run dev`

espindola:
When testing this in deno mode, it hits https://github.com/ije/esm.sh/issues/363.

It is interesting that deno produces different errors for x-typescript-types and `@deno-types`. My understanding is that they should be equivalent. I will investigate where the difference is coming from and it is a feature or a bug.

A long standing TODO is to refactor the deno code into a deno-tsc crate and use that both in deno and chiselstrike. I guess I should bump the priority of that item.

espindola:
Also reported https://github.com/denoland/deno/issues/15132

penberg:
Nevera on Discord says that this issue blocks `axios` and `jsonwebtokens` too, which makes this a high priority.

I see we have an ticket open on `jsonwebtokens` (although I thought it was stale until now): https://github.com/chiselstrike/chiselstrike/issues/1051

penberg:
Here's a simple reproducer for axios:

```
import axios from 'axios';

export default async function (req: Request): Promise<string> {
    return await axios.get("www.google.com");
}
```

```
Unhandled exception: Error: Promise rejected without a handler: Error: Dynamic require of "util" is not supported
    at file:///worker.js:6:15
could not import endpoint code into the runtime

Caused by:
    Error: Dynamic require of "util" is not supported
        at file:///dev/endpoints/hello:12:9
        at node_modules/combined-stream/lib/combined_stream.js (file:///dev/endpoints/hello:1751:16)
[...]

PR #1783, opened 2022-09-21
Import polyfills in place of Node standard modules

honzasp:
In short, this PR implements `require('buffer')`.

---

In more detail, this PR:

- Embeds the Deno standard library under internal URL `chisel://deno-std/`
- Adds a crate `deno_std`, which transpiles the library using swc (wrapped in `deno_ast` for convenience). I decided to use swc instead of our `tsc_compile`, because swc is faster, simpler and in the future, I would like to use swc in `chisel apply` anyway.
- The sources of the Deno standard library are vendored as a submodule in the repo. We should keep `third_party/deno_std` in sync with `third_party/deno`.
- Resolves Node global imports such as `'buffer'` into Node polyfills from the Deno standard library.
- Implements fake env and cwd, because npm packages often try to read from them.

### DIFF
--- a/cli/tests/integration_tests/rust_tests/imports.rs
+++ b/cli/tests/integration_tests/rust_tests/imports.rs
@@ -48,12 +48,12 @@ async fn deno_land_module(mut c: TestContext) {
     );
 }
 
-#[self::test(modules = Both)]
-async fn builtin_std(c: TestContext) {
+#[self::test(modules = Node)]
+async fn builtin_deno_std(c: TestContext) {
     c.chisel.write(
         "routes/semver.ts",
         r##"
-        import semver from 'chisel://std/semver/mod.ts';
+        import * as semver from 'chisel://deno-std/semver/mod.ts';
         export default async function () {
             return semver.gt("1.13.1", "1.9.9");
         }"##,
@@ -62,7 +62,7 @@ async fn builtin_std(c: TestContext) {
     c.chisel.write(
         "routes/node_buffer.ts",
         r##"
-        import { Buffer } from 'chisel://std/node/buffer.ts';
+        import { Buffer } from 'chisel://deno-std/node/buffer.ts';
         export default async function () {
             return Buffer.alloc(100).byteLength;
         }"##,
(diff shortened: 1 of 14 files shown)


penberg:
Fixes https://github.com/chiselstrike/chiselstrike/issues/1530, right?

honzasp:
> Fixes #1530, right?

Yes, it should.

cli/tests/integration_tests/rust_tests/imports.rs
```diff
@@ -48,3 +47,95 @@ pub async fn deno_land_module(mut c: TestContext) {
         "test    foo"
     );
 }
+
+#[self::test(modules = Node)]
```
penberg, line 51: Is this `Node` mode on purpose? The URL-based imports are usually only in Deno mode, no?

server/src/worker.rs
```diff
@@ -66,6 +66,12 @@ pub struct WorkerState {
     ///
     /// To wait on this channel, we temporarily move out of the `Option`.
     pub job_rx: Option<mpsc::Receiver<VersionJob>>,
+
+    /// Fake environment variables.
+    ///
+    /// We don't let the user code read the actual environment variables of this process, but we
+    /// implement `set_env` and `get_env` using this map.
+    pub fake_env: HashMap<String, String>,
```
penberg, line 74: What's the reason for faking these? Deno sandboxes environment variable access anyway so why not use that mechanism?

cli/tests/integration_tests/rust_tests/imports.rs
honzasp: This test is run only in Node mode because Deno mode typechecks the code, and `tsc` does not know how to deal with the `chisel://` imports. I added this test only to check an intermediate step in the implementation, it uses an "internal" API, so perhaps it should not be here at all.

server/src/worker.rs
penberg: @honzasp Ok, makes sense. The issue I am worried about is management of the environment variables, but I guess as long as we document this behaviour (heads up @CodingDoug), it's fine.

cli/tests/integration_tests/rust_tests/imports.rs
honzasp: Well, the "Node mode" is nothing but a set of kludges _on top_ of vanilla Deno, so all code that works for Deno should also work for Node. (In fact, we probably should not have two modes, but a single mode that supports the Node module resolution and the Node polyfills.)

penberg (approved):

PR #1784, opened 2022-09-21
Propagate TypeSystem to JavaScript and add Date support.

BearLemma:
This PR can perhaps be simplified if we merge https://github.com/chiselstrike/chiselstrike/pull/1780 first :thinking: 
That would completely remove the necessity to propagate the type system to TypeScript... Although it feels kinda beneficial anyway because of the proper initialization of nested entities. But it still might be better to decouple the two.

### DIFF
--- a/cli/tests/integration_tests/rust_tests/model_date.rs
+++ b/cli/tests/integration_tests/rust_tests/model_date.rs
@@ -1,0 +1,112 @@
+use crate::framework::prelude::*;
+
+#[chisel_macros::test(modules = Deno)]
+pub async fn store_and_load(c: TestContext) {
+    c.chisel.write(
+        "models/dated.ts",
+        r#"
+        import { ChiselEntity } from "@chiselstrike/api"
+        export class Dated extends ChiselEntity {
+            date: Date;
+        }"#,
+    );
+    c.chisel.write(
+        "routes/store.ts",
+        r#"
+        import { Dated } from "../models/dated.ts";
+        export default async function chisel(req: Request) {
+            await Dated.create({
+                date: Date.parse('01 Sep 2022 12:13:14 GMT'),
+            });
+        }
+        "#,
+    );
+    c.chisel.write(
+        "routes/read.ts",
+        r#"
+        import { Dated } from "../models/dated.ts";
+        export default async function chisel(req: Request) {
+            let dated = (await Dated.findOne({}))!;
+            return dated.date.toUTCString();
+        }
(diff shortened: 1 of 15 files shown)


server/src/types/mod.rs
```diff
@@ -28,6 +29,7 @@ impl Type {
             Type::Float => "number".to_string(),
             Type::String => "string".to_string(),
             Type::Boolean => "boolean".to_string(),
+            Type::JsDate => "jsDate".to_string(),
```
honzasp, line 32: I wonder, for all other types, the returned string is TypeScript-compatible, so perhaps this should also return `"Date"`? (the `"jsDate"` string is used in many other places, I don't know which occurences would also have to change).

cli/tests/integration_tests/rust_tests/model_date.rs
```diff
@@ -0,0 +1,111 @@
...
+            return dated.date.toUTCString();
+        }
+        "#,
+    );
+    c.chisel.apply_ok().await;
+    c.chisel.post_json("/dev/store", json!({})).await;
+    assert_eq!(
+        c.chisel.get_text("/dev/read").await,
+        "Thu, 01 Sep 2022 12:13:14 GMT"
+    );
```
honzasp, line 39: Did you consider using `c.chisel.get("/dev/read").send().await.assert_text("...")`? It will provide better error messages, for example if the HTTP request returns an error.

cli/tests/integration_tests/rust_tests/model_date.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
```
honzasp, line 41: I suggest that we also test whether `Date[]` works correctly.

honzasp (commented):
Looks very nice! :) I have some comments, but none of them are critical, so feel free to ignore/disagree. I do not understand the code in `server::datastore`, so I cannot verify the correctess of your changes here.

server/src/types/mod.rs
honzasp: Ok, this brings the question, what is the name even used for?

cli/tests/integration_tests/rust_tests/model_date.rs
honzasp: Uhhh, ok, now I see the special handling of `map_array_type` vs. `map_type` in `cli::ts`; yeah, let's fix that in another PR :)

PR #1785, opened 2022-09-21
Support for Array buffer

BearLemma:
This PR brings the support for `ArrayBuffer` type as Entity field.
Some notes:

- TS is perfectly fine assigning `ArrayBufferView` to `ArrayBuffer`. Luckily I was able to handle that case as well so no problem :relaxed: It's really unexpected behavior though...
- In CRUD, the ArrayBuffer is encoded with `base64`.
- The base64 encoding is handled well on the way out, but in `POST/PUT` endpoints, it's a bit tricky as there, the json goes through `Entity.build(...)` which meant I had to add ugly hack to the `mergeSourceIntoEntity` function. 
- The `mergeSourceIntoEntity` is becoming rather hairy and I think it will be necessary to refactor it in consequent PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -611,6 +611,7 @@ dependencies = [
  "js-sys",
  "num-integer",
  "num-traits",
+ "serde",
  "time 0.1.44",
  "wasm-bindgen",
  "winapi",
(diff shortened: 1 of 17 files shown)


penberg (commented):
lgtm, but looks like @honzasp has some outstanding review comments so not approving

honzasp (approved):

PR #1786, opened 2022-09-22
prometheus metrics

glommer:
This PR adds prometheus metrics to chiseld. An external prometheus server can then scrape these metrics.

Design decisions:
- There are two types of metrics: core and app. Core metrics are for core things in chiseld, like endpoints being called, and other implementation details that we may want to track.

  App metrics are defined by the user. So as not to force a metrics registration step, we register all user metrics with the same name (app_metric), and specify them by a tag. At this moment app metrics cannot have their own tags.

- Metrics are exposed through the internal health server, and not through the API.

[] Add more core metrics, like latency histograms. [] Expose other forms of metrics to the app, like gauge and histogram. [] Add tests.


server/src/metrics.rs
```diff
@@ -0,0 +1,79 @@
...
+        let mut buffer = vec![];
+        let encoder = TextEncoder::new();
+        let metric_families = self.reg.gather();
+        encoder.encode(&metric_families, &mut buffer).unwrap();
+
+        String::from_utf8(buffer).unwrap()
+    }
+}
+
+pub(crate) static GLOBAL_METRICS: Lazy<Metrics> = Lazy::new(Metrics::new);
```
honzasp, line 79: Instead of a global variable, it might be cleaner to add this object to the `Server` struct.

api/src/run.ts
```diff
@@ -49,9 +49,10 @@ export default async function run(
             break;
         } else if (job.type == "http") {
             const httpResponse = await handleHttpRequest(router, job.request);
-            opSync("op_chisel_http_respond", job.responseTxRid, httpResponse);
+            opSync("op_chisel_http_respond", job.responseTxRid, { response: httpResponse, uri: job.request.uri });
```
honzasp, line 52: I would suggest using a dedicated op to log the response (like you do with Kafka events below), instead of coupling the metric recording to the job machinery.

server/src/ops/mod.rs
```diff
@@ -83,3 +85,15 @@ fn op_chisel_is_debug(state: &mut deno_core::OpState) -> bool {
 fn op_format_file_name(file_name: String) -> Result<String> {
     Ok(file_name)
 }
+
+#[deno_core::op]
+fn op_chisel_app_counter(name: String) -> Result<()> {
```
honzasp, line 90: To make the syntax a bit sweeter, the function can simply return `()`, so you don't have to write the `Ok(())` :)

honzasp: It might also be better to make the ops consistent with each other, so perhaps this would be named `op_chisel_count_app` or something?

server/src/metrics.rs
glommer: is it available globally, from all contexts? Happy to look into it.

server/src/ops/mod.rs
glommer: Thanks, I thought a result was mandatory for the api.

"count app" makes me believe we are counting apps =) This is a counter that the app uses.

honzasp: I agree, I would in fact provide a generic API to increment arbitrary metric with arbitrary tags, but I didn't want to be too radical :)

What about `op_chisel_incr_app_counter`/`op_chisel_incr_kafka_event`/`op_chisel_incr_http_request`?

PR #1788, opened 2022-09-22
cli/apply: Run "esbuild" without "npx"

penberg:
The "esbuild" program is a native binary. Let's run it without "npx" to eliminate the Node.js dependency in "chisel apply".

### DIFF
--- a/cli/src/cmd/apply/node.rs
+++ b/cli/src/cmd/apply/node.rs
@@ -146,7 +146,7 @@ pub(crate) async fn apply(
         },
     ];
 
-    let bundler_output = npx("esbuild", &bundler_args)?
+    let bundler_output = esbuild(&bundler_args)?
         .wait_with_output()
         .await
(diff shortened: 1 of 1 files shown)


cli/src/cmd/apply/node.rs
```diff
@@ -162,6 +162,17 @@ pub(crate) async fn apply(
     Ok((modules, index_candidates))
 }
 
+fn esbuild<A: AsRef<OsStr>>(args: &[A]) -> Result<tokio::process::Child> {
```
honzasp, line 165: Do you think it would be useful to merge `esbuild` and `npx` into a single function (`run()`? `command()`? `execute()`?) to make this a bit drier?

glommer: for a follow up if Pekka wants it.

glommer (approved):

PR #1792, opened 2022-09-23
Upgrade to Deno v1.25.4

penberg:


penberg:
The `missing-await` test doesn't return 500, but instead now returns:

```
Error in GET /dev/find: Error: Cannot commit a transaction because there is an operation in progress that uses this transaction
    at async handleHttpRequest (chisel://api/http.ts:32:9)
    at async run (chisel://api/run.ts:21:34)
    at async chisel://api/main.js?version=dev&worker=0:11:1
Error when rolling back transaction: Error: Cannot rollback a transaction because no transaction is in progress
```

penberg:
The missing await test issue was resolved by translating it into Rust in commit 668b58e420c2c47a0139060dfc49af39cbe8003c

PR #1798, opened 2022-09-26
Implement support for Id field type.

BearLemma:
Adds support for `Id` type which can be used to model relationships.

### DIFF
--- a/api/src/datastore.ts
+++ b/api/src/datastore.ts
@@ -877,6 +877,40 @@ export class ChiselEntity {
     }
 
     /**
+     * Returns a single entity of type T that is identified by `entityId`.
+     *
+     * If no such entity is found, an exception is thrown.
+     */
+    static async byId<T extends ChiselEntity>(
+        this: typeof ChiselEntity & { new (): T },
+        entityId: Id<T>,
+    ): Promise<T> {
+        const entity = await this.findById<T>(entityId);
+        if (entity === undefined) {
+            throw Error(`failed to find entity with id '${entityId}'`);
+        }
+        return entity;
+    }
+
+    /**
+     * Returns a single entity of type T that is identified by `entityId`.
+     *
+     * If no such entity is found, returns undefined.
+     */
+    static async findById<T extends ChiselEntity>(
+        this: { new (): T },
+        entityId: Id<T>,
+    ): Promise<T | undefined> {
+        const it = chiselIterator<T>(this).filter(
+            { id: entityId } as Partial<T>,
+        ).take(1);
+        for await (const value of it) {
+            return value;
+        }
+        return undefined;
+    }
+
+    /**
      * Deletes all entities that match the `restrictions` object.
      *
      * @example
@@ -1163,7 +1197,7 @@ function mergeSourceIntoEntity(
                 `field ${field.name} of entity ${entityName} is ${typeName}, but provided value ${fieldValue} is of type ${typeof fieldValue}`,
             );
         };
-        if (field.type.name == "string") {
+        if (field.type.name == "string" || field.type.name == "entityId") {
             if (typeof fieldValue == "string") {
                 target[field.name] = fieldValue;
             } else {
@@ -1226,3 +1260,5 @@ function mergeSourceIntoEntity(
         }
     }
 }
+
+export type Id<Entity extends ChiselEntity> = Entity["id"];
--- a/cli/src/ts.rs
+++ b/cli/src/ts.rs
@@ -17,7 +17,7 @@ use swc_ecma_ast::{
     TsEntityName, TsKeywordTypeKind, TsType, TsTypeAnn,
 };
 use swc_ecma_parser::{lexer::Lexer, Parser, StringInput, Syntax, TsConfig};
-use swc_ecmascript::ast::{self as swc_ecma_ast};
+use swc_ecmascript::ast::{self as swc_ecma_ast, TsTypeRef};
 use swc_ecmascript::parser as swc_ecma_parser;
 
 impl FieldDefinition {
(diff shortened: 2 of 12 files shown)


api/src/datastore.ts
```diff
@@ -933,6 +934,19 @@ export class ChiselEntity {
         return undefined;
     }
 
+    static async byId<T extends ChiselEntity>(
+        this: { new (): T },
+        entity_id: Id<T>,
+    ): Promise<T | undefined> {
```
honzasp, line 940: I suggest that we provide two variants of this method, first that returns `T` or throws, and second that returns `T | undefined`. In my opinion, the first variant is going to be much more common, because there are going to be few cases where you look for an entity by its id and expect that it might not exist (for the same reason that there are few cases where you have a pointer and expect that the pointed-to object might not exist).

So I propose that the short `byId()` return `T`, and a longer variant, `byIdOptional()` or `findById()` or `getById()` or whatever,  return `T | undefined`.

cli/src/ts.rs
```diff
@@ -149,6 +152,30 @@ fn map_type(handler: &Handler, x: &TsType) -> Result<TypeEnum> {
     }
 }
 
+fn map_entity_id(handler: &Handler, tr: &TsTypeRef) -> Result<TypeEnum> {
+    let type_params = &tr
+        .type_params
+        .as_ref()
+        .context("Id field declared without Entity parameter")?
```
honzasp, line 159: I would reword this message to more accurately reflect the error, something like "Type `Id` requires type arguments".

cli/src/ts.rs
```diff
@@ -149,6 +152,30 @@ fn map_type(handler: &Handler, x: &TsType) -> Result<TypeEnum> {
...
+        .params;
+    ensure!(
+        type_params.len() == 1,
+        "Unexpected number of Id generic parameters"
```
honzasp, line 163: Similar to above, I would reword it something like "Type `Id` requires exactly one type argument", to be similar to error messages produced by `tsc`.

cli/src/ts.rs
```diff
@@ -149,6 +152,30 @@ fn map_type(handler: &Handler, x: &TsType) -> Result<TypeEnum> {
...
+    if let TsType::TsTypeRef(tr) = &**param {
+        if let TsEntityName::Ident(entity_id) = &tr.type_name {
+            let entity_name = ident_to_string(entity_id);
+            return Ok(TypeEnum::EntityId(entity_name));
+        }
+    }
+    Err(swc_err(
+        handler,
+        tr,
+        "failed to parse entity name out of Id field",
```
honzasp, line 175: What about "The type argument of type `Id` must refer to an entity"?

cli/tests/integration_tests/rust_tests/model_ids.rs
```diff
@@ -0,0 +1,120 @@
...
+            const trial = await Book.findOne({title: "The Trial"});
+            const author = await Author.byId(trial!.author);
+            return author!.name;
+        }"#,
+    );
+
+    c.chisel.apply_ok().await;
+    c.chisel.post("/dev/store").send().await.assert_ok().text();
+    c.chisel
+        .get("/dev/get_joke_author")
```
honzasp, line 50: Probably `get_trial_author`? :)

cli/tests/integration_tests/rust_tests/model_ids.rs
```diff
@@ -0,0 +1,120 @@
...
+        "models/models.ts",
+        r#"
+        import { ChiselEntity, Id } from "@chiselstrike/api";
+
+        export class Book extends ChiselEntity {
+            title: string;
+            author: Id<Author>;
+        }"#,
+    );
+    c.chisel.apply_err().await;
```
honzasp, line 119: I think we should assert that it failed with an expected error, rather than for some unexpected reason.

server/src/ops/type_system.rs
```diff
@@ -57,6 +57,10 @@ enum SimpleTypeId {
         #[serde(rename = "entityName")]
         entity_name: String,
     },
+    EntityId {
+        #[serde(rename = "entityId")]
```
honzasp, line 61: Did you mean `rename = "entityName"`?

server/src/types/mod.rs
```diff
@@ -20,6 +20,7 @@ pub enum Type {
     Boolean,
     JsDate,
     Entity(Entity),
+    EntityId { entity_name: String },
```
honzasp, line 23: It's a bit weird when one enum mixes tuple variants `Entity(Entity)` and struct variants `EntityId { entity_name: String }`, do you think that `EntityId(String)` would be unclear?

honzasp (commented):
Looks ok, I have just some minor comments, nothing really important.

MarinPostma (commented):
lgtm

MarinPostma:
I'm thinking that you could add a test that ensure that an entity `A` cannot be queried with an `Id<B>`?

MarinPostma:
right, but that defeats the purpose of typed ids a bit

cli/tests/integration_tests/rust_tests/model_ids.rs
```diff
@@ -0,0 +1,178 @@
...
+
+#[self::test(modules = Deno)]
+pub async fn basic(c: TestContext) {
+    write_models(&c.chisel);
+    c.chisel.write(
+        "routes/store.ts",
+        r#"
+        import { Book, Author } from '../models/models.ts';
+        export default async function () {
+            const kafka = await Author.create({
```
glommer, line 27: are we going to support RedPanda as well ?

honzasp (approved):

PR #1810, opened 2022-09-29
Fix Date's storage and replace `mergeDeep` with proper reflection-based building

BearLemma:
The fix seems to break the test which means that the `EntityValue::to_v8` method doesn't work as it should.
It looks that when the `v8::Value` of `Date` gets to our function, it's just a plain Object and we can't tell it apart by `is_date` method :(.

The reason why it worked so far is that because of CRUD - JSON we must accept float as well from there.

The fix consists of replacing all `mergeDeep` with `buildEntity` and `mergeIntoEntity` which are datastore functions leveraging the recently introduced `typeSystem` data to reflect over the types that are being built. Note that the functions are more or less just a copy of the `populateEntity` method, that was already there.

### DIFF
--- a/api/src/datastore.ts
+++ b/api/src/datastore.ts
@@ -2,7 +2,7 @@
 
 import { crud } from "./crud.ts";
 import type { RouteMap } from "./routing.ts";
-import { mergeDeep, opAsync, opSync } from "./utils.ts";
+import { opAsync, opSync } from "./utils.ts";
 import { SimpleTypeSystem, TypeSystem } from "./type_system.ts";
 
 /**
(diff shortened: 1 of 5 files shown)


api/src/datastore.ts
```diff
@@ -1158,3 +1102,78 @@ export async function loggedInUser(): Promise<AuthUser | undefined> {
...
+                continue;
+            }
+            const fieldValue = source[field.name];
+            switch (field.type.name) {
+                case "string":
+                case "number":
+                case "boolean":
+                case "array":
+                    target[field.name] = fieldValue;
+                    break;
```
honzasp, line 1142: Would it make sense to validate or convert these types, rather than simply accepting any value?

api/src/datastore.ts
```diff
@@ -1158,3 +1102,78 @@ export async function loggedInUser(): Promise<AuthUser | undefined> {
...
+                    if (fieldValue === undefined || fieldValue === null) {
+                        if (field.isOptional) {
+                            target[field.name] = fieldValue;
+                            continue;
+                        } else {
+                            throw new Error(
+                                `field ${field.name} of entity ${entityName} is not optional but backend returned no value`,
+                            );
+                        }
+                    }
```
honzasp, line 1158: Why do we perform the optionality check only for type `"entity"`, but not for other types?

api/src/datastore.ts
```diff
@@ -1158,3 +1102,78 @@ export async function loggedInUser(): Promise<AuthUser | undefined> {
...
+                    );
+                    break;
+                case "entity":
+                    if (fieldValue === undefined || fieldValue === null) {
+                        if (field.isOptional) {
+                            target[field.name] = fieldValue;
+                            continue;
+                        } else {
+                            throw new Error(
+                                `field ${field.name} of entity ${entityName} is not optional but backend returned no value`,
```
honzasp, line 1155: This error message might be confusing, because the `source` might also be provided by the user, when calling `ChiselEntity.create`.

honzasp (commented):
Very nice! Just a few comments; we might do more validation in `mergeIntoEntity()`.

honzasp (approved):

honzasp (dismissed):

honzasp (dismissed):

honzasp (approved):

PR #1815, opened 2022-09-30
Rewrite optional-field.deno test to Rust.

BearLemma:
The rewrite omits the last part of the test:

```
## Correctly populate from optional fields
$CHISEL apply --version=v2
# CHECK: Applied:
$CHISEL populate --version=v2 --from=dev
$CURL $CHISELD_HOST/v2/sum
# CHECK: HTTP/1.1 200 OK
# CHECK: the sum is: 1100;
```

Which is testing population of optional fields. This is because from what I've seen in the `populate.deno` file, this seems to be covered by populating optional `ceo` field. @dekimir as you are the author of the test, are you ok with this, or should I add an explicit population test as well?


honzasp (commented):
Looks reasonable; not formally approving before Dejan has a look, though :)

The tests `empty_and_null`, `number`, `string`, `boolean` and `array` are quite repetitive, maybe it would be possible to abstract them into a function parameterized by the type name and JSON values?

(I'm a bit uneasy that the test file is so long, even though it does not test a lot of functionality.)

honzasp:
> Yeah I had it all in a macro, but it was weird and didn't work for array and Entity... This looked like the best compromise between length and readability.

Oh yeah, macro sounds nasty; would not a plain function work? On a cursory look, the only difference between `number` and `string` seems to be in the type name given to `write_files()` and the JSON values given to requests and expected in responses.

honzasp (approved):
There is a typo in `run_primitve_test`, but meh :)

PR #1826, opened 2022-10-04
server: Fix typo in a comment

haaawk:


server/src/http.rs
```diff
@@ -187,7 +187,7 @@ async fn handle_version_request(
     // even if `send()` returns an `Ok`, it does not in fact guarantee that the job is received or
     // processed, so we _still_ must handle the case when the job is dropped ...
     let _: Result<_, _> = job_tx.send(job).await;
-    // ... which happens here: when the `job` is dropped, `job.response_tx` is also dropped, so the
+    // ... which happens here: when the `job` is dropped, `job.response_rx` is also dropped, so the
```
MarinPostma, line 190: I think that this is correct, the job referred to here is the job from like 186. When the job get dropped, its tx gets dropped as well, and line 192 fails

haaawk: Right. I've missed that. Thanks!

PR #1831, opened 2022-10-07
cli: Clean up FileRouteMap::add_route and its tests

haaawk:
Don't panic in add_route but return Result<()> instead. Use "" instead of r#""# in tests.

Signed-off-by: Piotr Jastrzebski <piotr@chiselstrike.com>

### DIFF
--- a/cli/src/routes.rs
+++ b/cli/src/routes.rs
@@ -13,14 +13,15 @@ pub(crate) struct FileRouteMap {
 }
 
 impl FileRouteMap {
+    #[allow(unused_must_use)]
     fn add_route(
         &mut self,
         file_path: PathBuf,
         path_pattern: String,
         legacy_file_name: Option<String>,
-    ) {
+    ) -> Result<()> {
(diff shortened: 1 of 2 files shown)


cli/src/routes.rs
```diff
@@ -13,12 +13,13 @@ pub(crate) struct FileRouteMap {
...
+    #[allow(unused_must_use)]
     fn add_route(
         &mut self,
         file_path: PathBuf,
         path_pattern: String,
         legacy_file_name: Option<String>,
-    ) {
+    ) -> Result<()> {
         if fs::metadata(&file_path)
             .unwrap_or_else(|_| panic!("Cannot fetch {} metadata", &file_path.display()))
```
dekimir, line 24: But, but... this still panics?!

haaawk: Yeah. How stupid is that :). Sorry about this brain fart. Uploaded a newer version where this is fixed.

dekimir (approved):

PR #1835, opened 2022-10-12
TS api: runtime check argument type for delete

MarinPostma:
This PR checks that the type of the argument passed to the `restrictionsToFilterExpr` is in fact an object. If it wasn't it would wipe out the database.


api/src/datastore.ts
```diff
@@ -1049,6 +1049,9 @@ export class ChiselEntity {
 function restrictionsToFilterExpr<T extends ChiselEntity>(
     restrictions: Partial<T>,
 ): Record<string, unknown> | undefined {
+    if (typeof restrictions != "object") {
+        throw `expected object, but found ${typeof restrictions} instead`;
```
honzasp, line 1053: I think we should throw `Error`-s, not strings.

PR #1837, opened 2022-10-13
ChiselStrike adapter for NextAuth.js

penberg:
This adds a ChiselStrike adapter for NextAuth.js. The adapter is based on the work Dejan did earlier:

    https://github.com/nextauthjs/next-auth/pull/4573

The package is standalone and, therefore, does not require NextAuth.js to integrate our work into their tree for our users to use this.

Please note that although you can't run the tests in the tree, they do pass if you integrate the package with the NextAuth.js repository and enable the "workspace" tagged dependencies (that only pnpm understands).


barbieri (dismissed):
I didn't re-review the lib/adapter part, but package wise looks good

packages/chiselstrike-next-auth/README.md
```diff
@@ -0,0 +1,50 @@
+# ChiselStrike Adapter for NextAuth.js
+
+## Getting Started:
+
+To use the adapter in your application, do the following:
```
dekimir, line 5: "do" is ambiguous here.  Please explain which file this is and where it goes.

packages/chiselstrike-next-auth/README.md
```diff
@@ -0,0 +1,50 @@
...
+import GitHubProvider from "next-auth/providers/github";
+import NextAuth from "next-auth";
+import { ChiselAdapter } from "@chiselstrike/next-auth";
+
+export const authOptions = {
+    adapter: ChiselAdapter({
+        url: "http://localhost:8080/__chiselstrike/auth/",
+        secret: "1234",
+    }),
+    providers: [
```
dekimir, line 17: This is irrelevant to the adapter.  I'd suggest replacing it with a placeholder.

packages/chiselstrike-next-auth/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+```typescript
+import GitHubProvider from "next-auth/providers/github";
+import NextAuth from "next-auth";
+import { ChiselAdapter } from "@chiselstrike/next-auth";
+
+export const authOptions = {
+    adapter: ChiselAdapter({
+        url: "http://localhost:8080/__chiselstrike/auth/",
+        secret: "1234",
```
dekimir, line 15: Values here should be abstract.

packages/chiselstrike-next-auth/README.md
```diff
@@ -0,0 +1,50 @@
...
+To use the adapter in your application, do the following:
+
+```typescript
+import GitHubProvider from "next-auth/providers/github";
+import NextAuth from "next-auth";
+import { ChiselAdapter } from "@chiselstrike/next-auth";
+
+export const authOptions = {
+    adapter: ChiselAdapter({
+        url: "http://localhost:8080/__chiselstrike/auth/",
```
dekimir, line 14: The adapter should add `__chiselstrike/auth` automatically, as this is an internal detail that can change.

packages/chiselstrike-next-auth/README.md
```diff
@@ -0,0 +1,50 @@
...
+export default NextAuth(authOptions);
+```
+
+For local development, define the auth secret in `.env` file of ChiselStrike
+project:
+
+```
+{ "CHISELD_AUTH_SECRET" : "1234" }
+```
+
```
dekimir, line 33: Why did you drop the part about the `ChiselUID` header?

packages/chiselstrike-next-auth/jest.config.js
```diff
@@ -0,0 +1,11 @@
+/*
+ * For a detailed explanation regarding each configuration property and type check, visit:
+ * https://jestjs.io/docs/en/configuration.html
+ */
+module.exports = {
+    ...require("@next-auth/adapter-test/jest/jest-preset"),
+    // // Indicates whether the coverage information should be collected while executing the test
```
dekimir, line 7: Let's remove these commented-out comments.  They're not helpful here.

packages/chiselstrike-next-auth/lib/adapter.ts
```diff
@@ -0,0 +1,232 @@
+export function ChiselAdapter(options: { url: string; secret: string }) {
+    const headers = { "ChiselAuth": options.secret };
+    const url = options.url;
+    return {
+        async createUser(data) {
```
dekimir, line 5: Why isn't the parameter typed?

packages/chiselstrike-next-auth/lib/adapter.ts
```diff
@@ -0,0 +1,232 @@
...
+            const user = JSON.stringify(data);
+            const response = await fetch(`${url}users`, {
+                method: "POST",
+                headers,
+                body: user,
+            });
+            return toUser(await response.json());
```
dekimir, line 12: What if response status isn't OK?

dekimir (changes requested):
This differs significantly from the existing adapter PR, and I'm not sure why.  Are all the changes really necessary?

packages/chiselstrike-next-auth/lib/adapter.ts
```diff
@@ -0,0 +1,232 @@
...
+            return await fetchUserById(url, headers, account.userId);
+        },
+        async updateUser(updatedUser) {
+            const user = await fetchUserById(url, headers, updatedUser.id);
+            if (!user) {
+                return;
+            }
+            Object.assign(user, updatedUser);
+            await fetch(`${url}users/${user.id}`, {
+                method: "PUT",
```
dekimir, line 39: This should be PATCH.

packages/chiselstrike-next-auth/lib/adapter.ts
```diff
@@ -0,0 +1,232 @@
...
+    );
+    if (!response.ok) {
+        return null;
+    }
+    const json = await response.json();
+    const sessions = json?.results;
+    if (!sessions?.length) {
+        return null;
+    }
+    return toSession(sessions[0]);
```
dekimir, line 163: This pattern (get `.results`, ensure `[0]` exists, return `[0]`) repeats at least four times.  Consider DRYing.

packages/chiselstrike-next-auth/lib/adapter.ts
penberg: I want to make all of them `PATCH`, but it doesn't work in the general case: https://github.com/chiselstrike/chiselstrike/issues/1845 Therefore, I use `PUT` everywhere.

packages/chiselstrike-next-auth/lib/adapter.ts
penberg: It indeed does, but that's a general problem with consuming a ChiselStrike REST API so the fix should be in a client library, not this adapter.

penberg:
This is a rewrite, not an adaptation. The main difference is using `fetch()` directly as per outstanding review comments here: https://github.com/nextauthjs/next-auth/pull/4573#discussion_r906810184

(And as I mentioned you privately already, I started the rewrite because I couldn't get the adapter to run. But as it turns out, the problem was not in the original adapter, but the way I used `Credentials` provider, but now I already had 90% of the adapter rewritten.)

packages/chiselstrike-next-auth/lib/adapter.ts
penberg: I just followed the style of the existing adapters that don't type them (don't know why). See, for example, https://github.com/nextauthjs/next-auth/blob/main/packages/adapter-firebase/src/index.ts.

penberg:
@dekimir You had bunch of comments on the documentation in `README.md` that I did not address, but I don't think are blocker for merging this.

To address the docs stuff, I think you, me, and @CodingDoug should sit down and figure out how to document this in `docs.chiselstrike.com` and then maybe replace the README with a link instead.

packages/chiselstrike-next-auth/lib/adapter.ts
```diff
@@ -0,0 +1,244 @@
+export function ChiselAdapter(options: { url: string; secret: string }) {
+    const headers = { "ChiselAuth": options.secret };
+    const url = options.url + "/__chiselstrike/auth/";
+    return {
```
haaawk, line 4: Following methods are not expected by [the Adapter interface](https://next-auth.js.org/tutorials/creating-a-database-adapter#required-methods) but are present in this implementation: `getAccount`, `getSession`, `getVerificationToken`. What's the purpose of their existence?

haaawk: Adapter interface -> https://github.com/nextauthjs/next-auth/blob/main/packages/next-auth/src/adapters.ts#L79

penberg: The extra helpers are used by the tests. If you look at other adapters in the source tree, there's often a separate `Client` type that's used by the `Adapter`. I tried that route, but then our `Adapter` needs to wrap most of the `Client` methods, which is why I ended up with this.

penberg: @haaawk That `Adapter` type certainly makes sense, but it's defined in the `next-auth/adapters` AFAICT, which is not a public package. IOW, our external adapter cannot depend on it.

haaawk: Cool. Thanks for the explanation.

packages/chiselstrike-next-auth/lib/adapter.ts
dekimir: This isn't the general case (nor is the other PUT, for that matter).  This will overwrite existing fields that `updatedUser` omits, which is incorrect behaviour.

dekimir:
> This is a rewrite, not an adaptation.

I'm not sure a rewrite is justified when we already have working code.  What is your objection to using the adapter from the next-auth PR verbatim?  It contains better documentation, more tests, more error handling, and more code reuse.

> The main difference is using `fetch()` directly as per outstanding review comments here: [nextauthjs/next-auth#4573 (comment)](https://github.com/nextauthjs/next-auth/pull/4573#discussion_r906810184)

That comment was never properly justified.  Following it would make our adapter different from most others while duplicating a bunch of code (or merging another class into the adapter, not sure which; my clarifying questions were ignored).

packages/chiselstrike-next-auth/lib/adapter.ts
penberg: @dekimir Looking it again, you're right that the other `PUT` can also be replaced with `PATCH` (I guess the code was different at some point), so fixed.

dekimir (approved):
Approving to make progress on a long list of tasks awaiting us.  We'll need documentation and testing work anyway, whatever the difference between this PR and the next-auth PR.

packages/chiselstrike-next-auth/lib/adapter.ts
dekimir: On second thought, I'm wrong in claiming this is incorrect behaviour: this is exactly like PATCH due to the previous `Object.assign()` line.  Sorry.

packages/chiselstrike-next-auth/lib/adapter.ts
```diff
@@ -0,0 +1,246 @@
...
+                provider,
+                providerAccountId,
+            );
+            if (!account) {
+                return null;
+            }
+            return await fetchUserById(url, headers, account.userId);
+        },
+        async updateUser(updatedUser) {
+            const user = await fetchUserById(url, headers, updatedUser.id);
```
dekimir, line 36: We probably don't need this fetch; PATCH will do the right thing all by itself, I suspect.

PR #1841, opened 2022-10-15
Add _createdAt and _updatedAt fields

glommer:
This PR adds _createdAt and _updatedAt fields. We will establish a rule, that we now check, that ChiselStrike internal entities should start with an underscore.

### DIFF
--- a/api/src/datastore.ts
+++ b/api/src/datastore.ts
@@ -679,6 +679,16 @@ export class ChiselEntity {
     /** UUID identifying this object. */
     id?: string;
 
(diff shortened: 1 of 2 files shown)


glommer:
Draft, because tests don't pass with nested entities.

@BearLemma, I have no idea why and would appreciate your help.

Draft PR is so that we can validate the approach early instead of me bouncing on it for days (which I may not have time to do) so I am tagging some folks that can potentially have feedback:

@penberg @BearLemma @CodingDoug @luisfvieirasilva @honzasp

PR #1849, opened 2022-10-17
filters full ts support

MarinPostma:
The goal of this pr is to support for more typescript inside of the policy filters.

Currently, the condition branches in the filter functions only accepted a single statement, and refused to compile if there was any more. The new implementation is smarter.

1. If a branch that would lead to a skip has more than one statement, then we abort optimization, and return a warning to the user saying that we failed to optimize. For example:

```typescript
export default {
    read: (person, ctx) => {
        if (person.name == "hello") {
            debug("hello");
            return Action.Skip;
        }
    }
}
```
 returns the following warning on compile:

 ```
warning: Could not optimize skip branch. Consider making the return statement the only statement in the skip branch block.
 --> file:///Users/mpostma/Documents/code/rust/chiselstrike/test/policies/Person.ts:4:13
  |
4 |             debug("hello");
  |             ^^^^^^^^^^^^^^^
 ```

 2. For branches that do not lead to a skip, there are no restrictions.

 On top of those improvement, this PR add extensive testing of the filter. It works by writing typescript policy code where all predicates in if-conditions are given variable name `p1, p2, ...`. The test function will generate all combination of booleans for these variables, and perform substribtution on the code. The skip_condition is then extracted, and compared to the output of running the code, exhaustively.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -655,15 +655,20 @@ version = "0.15.0-dev.0"
 dependencies = [
  "anyhow",
  "bimap",
+ "boa_engine",
  "clap 4.0.27",
  "indexmap",
+ "itertools 0.10.5",
+ "maplit",
  "petgraph 0.6.2",
  "quine-mc_cluskey",
+ "regex",
  "serde",
  "serde_json",
  "swc_atoms 0.3.1",
  "swc_common 0.25.0",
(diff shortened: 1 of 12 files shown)


MarinPostma:
@BearLemma I had this bad boy left behind for some time, commit history is _not great_, lmk if I can help you review that

PR #1850, opened 2022-10-18
Add `chisel generate` command

penberg:
This adds a `chisel generate` command that generates types that can be
used from client code to simplify accessing ChiselStrike endpoints.

Currently, the command generates types based on the entities. For
example, given the following entity:

```typescript
export class Message extends ChiselEntity {
    sender: string;
    body: string;
    timestamp: string;
}
```

The command generates the following types:

```typescript
export type Message = {
    id: string;
    sender: string;
    body: string;
    timestamp: string;
}
export type Results<T> = {
    next_page: string;
    prev_page: string;
    results: T[];
}
```

The `Message` type maps to the `Message` entity directly. The
`Results<T>` type is the output of a CRUD endpoint that returns multiple
entities.

In the future, if we change the endpoint API to have more types, this
command can generate a full-blown RPC-like client with proper typing.

### DIFF
--- a/cli/src/cmd/generate.rs
+++ b/cli/src/cmd/generate.rs
@@ -1,0 +1,53 @@
+use crate::proto::chisel_rpc_client::ChiselRpcClient;
+use crate::proto::{type_msg::TypeEnum, DescribeRequest};
+use anyhow::{anyhow, Result};
+use std::fs::File;
+use std::io::Write;
+use std::path::PathBuf;
+
+pub(crate) async fn cmd_generate(
+    server_url: String,
+    output: PathBuf,
+    version: String,
+) -> Result<()> {
+    let output = File::create(output)?;
+    let mut client = ChiselRpcClient::connect(server_url).await?;
+    let request = tonic::Request::new(DescribeRequest {});
+    let response = execute!(client.describe(request).await);
+    let version_def = response
+        .version_defs
+        .iter()
+        .find(|def| def.version_id == version);
+    if let Some(version_def) = version_def {
+        for def in &version_def.type_defs {
+            writeln!(&output, "export type {} = {{", def.name)?;
+            writeln!(&output, "    id: string;")?;
+            for field in &def.field_defs {
+                let field_type = field.field_type()?;
+                writeln!(
+                    &output,
+                    "    {}{}: {}{};",
+                    field.name,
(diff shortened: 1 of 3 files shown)


cli/src/cmd/generate.rs
```diff
@@ -0,0 +1,52 @@
...
+                    &output,
+                    "    {}{}: {}{};",
+                    field.name,
+                    if field.is_optional { "?" } else { "" },
+                    field_type,
+                    field
+                        .default_value
+                        .as_ref()
+                        .map(|d| if matches!(field_type, TypeEnum::String(_)) {
+                            format!(" = \"{}\"", d)
```
MarinPostma, line 36: I think this is more readable :)
```suggestion
                            format!(r#" = "{}""#, d)
```

cli/src/cmd/generate.rs
```diff
@@ -0,0 +1,52 @@
...
+    version: String,
+) -> Result<()> {
+    let output = File::create(output)?;
+    let mut client = ChiselRpcClient::connect(server_url).await?;
+    let request = tonic::Request::new(DescribeRequest {});
+    let response = execute!(client.describe(request).await);
+    for version_def in response.version_defs {
+        if version_def.version_id != version {
+            continue;
+        }
```
MarinPostma, line 20: I would do something like:

```
let version_def = response.version_defs.iter().find(|def| dev.version_id == id);
```
if this is `None`, I would return an error saying that there is not such version

cli/src/cmd/generate.rs
```diff
@@ -0,0 +1,52 @@
...
+        for def in &version_def.type_defs {
+            writeln!(&output, "export type {} = {{", def.name)?;
+            writeln!(&output, "    id: string;")?;
```
MarinPostma, line 23: It doesn't matter _that much_ here, but we are performing blocking IO in an async context. We should be wary of that.

MarinPostma (commented):
LGTM mod few nits

Ultimately, I think code gen should be done at the AST level, but this does not need to be done now

cli/src/cmd/generate.rs
penberg: Good point, but this is a CLI command so we don't care, I think.

Issue #1833, opened 2022-10-07
Add `.count()` method to cursor API


luisfvieirasilva:
Would be great to have `count()` method that returned only the number of elements that `toArray()` would return

Issue #999, opened 2022-03-23
pagination in API results (JSON) and .count()


mpdehaan:
pagination in the REST layer is important for client apps to avoid returning giant result sets over JSON and also making life easy on the caller when implementing things like shopping carts (page 1 of N, 987 total items)

To get this, it would be nice if pagination were supported.

One example of doing this is in a REST API that I really like is https://docs.djangoproject.com/en/4.0/topics/pagination/ ... 

we could possibly annotate (or policy or whatver) a TypeScript model collection as being paginated, so simple cases wouldn't have to do it, but from my experience with developers is they often don't think about pagination until someone yells at them, so I think it would make a great default.

you may allow them some query string control over page size but there should always be server side limits.

GitHub takes a different approach to pagination (HTTP headers) which is IMHO much less friendly.

it would be also nice if the API result set supported .count()

PR #1855, opened 2022-10-19
Introduce `count` aggregation to TS API

BearLemma:
Introduces `async count(): Promise<number>` method to our cursor API.
closes https://github.com/chiselstrike/chiselstrike/issues/999
closes https://github.com/chiselstrike/chiselstrike/issues/1833

One rather annoying thing is that I had to add `TypeId::Int64` to make Postgress happy. The reason is that `COUNT(*)` returns an integer and if we try to load it as `f64` via `sqlx` in Postgres, it blows up. It doesn't blow up for Sqlite for some reason...

And another thing is that the Datastore architecture is clearly being stretched very thin and I'm really looking forward to the new Datamodel PR which will have come up with a brand new abstraction that fits us better.


api/src/datastore.ts
```diff
@@ -385,6 +385,35 @@ class SortBy<T> extends Operator<T, T> {
...
+    }
+
+    apply(
+        iter: AsyncIterable<unknown>,
+    ): AsyncIterable<number> {
+        return {
+            [Symbol.asyncIterator]: async function* () {
+                let count = 0;
+                for await (const _ of iter) {
+                    count += 1;
```
haaawk, line 405: Would it be possible to just get the count from the DB?

MarinPostma: @BearLemma, @haaawk hmm we need to think about that. say I do:
```typescript
let cursor = Person.select(p => p.age > 32);
for await (const p of cursor) {
     console.log(p.name)
}
let count = await cursor.count();
```

what should count be?

honzasp (commented):
Looks reasonable, but I agree that it stretches the current datastore code past its breaking point.

honzasp (approved):

PR #1862, opened 2022-10-22
Upgrade to deno_std 0.160.0

penberg:
Sync up `deno_std` to Deno v1.26.2 as per:

https://raw.githubusercontent.com/denoland/dotland/main/versions.json


penberg:
@BearLemma As discussed privately, I think it's unlikely that `deno_std` upgrade affects the build process: it's more like that the Deno/Rust upgrade that is already in `main` broke it.

PR #1870, opened 2022-10-26
fix stream-db.deno test

MarinPostma:
This PR fixes the stream-db.deno test by making sure that all the writes are performed sequentially.

I found this when working on an implementation of `add_row` that had more than one `await` point. Awaiting the promises concurrently gives no guarantees that the insert will be performed in order.

### DIFF
--- a/cli/tests/integration_tests/lit_tests/stream-db.deno
+++ b/cli/tests/integration_tests/lit_tests/stream-db.deno
@@ -28,7 +28,7 @@ export default async function chisel(req: Request) {
     }
     await Promise.all(promises);
 
-    let cursor = Data.cursor();
+    let cursor = Data.cursor().sortBy("num");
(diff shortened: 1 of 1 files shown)


honzasp:
I'm not sure what this test is supposed to test, but the root of the problem is that the test depends on `Data.cursor()` returning the entities in order of `num`. Maybe a better fix would be to use `Data.cursor().sortBy("num")`?

MarinPostma:
One way on another, we have to make ordering deterministic, both approaches fix the issue

honzasp:
Well, I don't think that we guarantee that `Data.cursor()` produces data in insertion order, so making sure that the ordering is deterministic by explicit `sortBy()` might be the more robust fix.

MarinPostma:
ok fair :)

PR #1871, opened 2022-10-26
Prometheus

athoscouto:
This PR creates support for prometheus in chiseld with the following additions:

- It creates a Prometheus registry and exposes it through the internal HTTP server, at /metrics.
- It creates Deno ops that allow Typescript code to register and increment `f64` counters with arbitrary names and labels.
- It creates a typescript class that allow users to register and increment counters through typescript. It also exposes it through the `@chiselstrike/api` module.

### Differences with https://github.com/chiselstrike/chiselstrike/pull/1786:
- This allows typescript code to create arbitrary counters, instead of using a single one with different labels. We can consider adding a prefix to any counters created on TS. Another option would be using a prefix on all rust metrics and prevent TS code from using the same prefix.
- This PR does not create counters with the metrics object. This should be done in a subsequent PRs.
- Counters exposed to TS are `f64` instead of `u64`. This representation is closer to TS `number`, at cost of perf. If that turns out to be a problem, we can introduce an integer counter later.
- 
### Pending
- [ ] Add tests. Advices on ways to do it are welcome.


server/src/metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use once_cell::sync::Lazy;
+use prometheus::{
+    core::{AtomicF64, GenericCounter},
+    CounterVec, Encoder, IntGaugeVec, Opts, Registry, TextEncoder,
+};
+
+pub(crate) struct Metrics {
+    reg: Registry,
+    counters: RwLock<HashMap<String, CounterVec>>,
```
honzasp, line 14: Instead of `std::sync::RwLock`, you may consider using `parking_lot::RwLock`; they are claimed to be faster, and they don't use poisoning, so you don't need to check for errors when trying to lock them.

server/src/metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+        Ok(())
+    }
+
+    fn get_counter(&self, name: &str, labels: &[&str]) -> Result<GenericCounter<AtomicF64>> {
+        Ok(self
+            .counters
+            .read()
+            .map_err(|_| anyhow!("Failed to aquire metrics read lock."))?
+            .get(name)
+            .ok_or_else(|| anyhow!("Could not find metric counter {name}."))?
```
honzasp, line 70: Another nitpick: error messages in Rust usually don't end with a period

api/src/utils.ts
```diff
@@ -12,6 +12,68 @@ export function opAsync(
...
+    labels?: T[];
+}
+
+interface IncrementCounterByParams {
+    name: string;
+    labels: string[];
+    value: number;
+}
+
+type Labels<T extends string> = Record<T, string>;
```
honzasp, line 27: Can `T` be something else than `string`? Perhaps a `symbol`?

athoscouto: TIL about symbols :)

I'm trying to think about a case where we would want anything other than strings.
In the end all labels are transformed into text when we expose metrics through HTTP.

[Prometheus data model](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels) is not explicit about it, but my understanding is that it limits labels to strings (by saying they must match a regex).
So I think we should limit it to string. If that turns out to be the wrong approach, we can always extend it later.

server/src/metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub(crate) struct Metrics {
+    reg: Registry,
+    counters: RwLock<HashMap<String, CounterVec>>,
+}
+
+impl Metrics {
+    pub(crate) fn new() -> Metrics {
+        let gauge_opts = Opts::new("version", "Currently running version of chiseld");
+        let version = IntGaugeVec::new(gauge_opts, &["version"]).unwrap();
```
haaawk, line 20: Another potential panic, no?

haaawk: And all the other unwrap in this function

server/src/metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+        let encoder = TextEncoder::new();
+        let metric_families = self.reg.gather();
+        encoder.encode(&metric_families, &mut buffer).unwrap();
+
+        String::from_utf8(buffer).unwrap()
+    }
+
+    pub(crate) fn register(&self, name: &str, help: &str, labels: &[&str]) -> Result<()> {
+        let counter_opts = Opts::new(name, help);
+        let counter = CounterVec::new(counter_opts, labels).unwrap();
```
haaawk, line 44: panic

server/src/metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    // FIXME: stream directly to body?
+    pub(crate) fn gather(&self) -> String {
+        let mut buffer = vec![];
+        let encoder = TextEncoder::new();
+        let metric_families = self.reg.gather();
+        encoder.encode(&metric_families, &mut buffer).unwrap();
+
+        String::from_utf8(buffer).unwrap()
```
haaawk, line 39: panic

server/src/metrics.rs
```diff
@@ -0,0 +1,75 @@
...
+impl Metrics {
+    pub(crate) fn new() -> Metrics {
+        let gauge_opts = Opts::new("version", "Currently running version of chiseld");
+        let version = IntGaugeVec::new(gauge_opts, &["version"]).unwrap();
+        version
+            .with_label_values(&[env!("VERGEN_GIT_SEMVER_LIGHTWEIGHT")])
+            .set(1);
+
+        let reg = Registry::new();
+        reg.register(Box::new(version)).unwrap();
```
haaawk, line 26: panic

api/src/utils.ts
```diff
@@ -12,6 +12,68 @@ export function opAsync(
...
 
+interface CounterParams<T extends string = string> {
+    name: string;
+    help?: string;
+    labels?: T[];
+}
+
+interface IncrementCounterByParams {
+    name: string;
+    labels: string[];
```
haaawk, line 23: Would it make sense to make this optional since labels in CounterParams are optional?

haaawk:
LGTM in general

server/src/internal.rs
```diff
@@ -25,6 +25,11 @@ fn response(body: &str, status: u16) -> Result<Response<Body>> {
         .unwrap())
 }
 
+fn collect_metrics() -> Result<Response<Body>> {
+    let s = crate::metrics::GLOBAL_METRICS.gather();
+    Ok(Response::builder().body(Body::from(s)).unwrap())
```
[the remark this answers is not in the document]
honzasp: It would panic if the response was invalid, for example if you specified an HTTP header with `\r\n` inside. I'd say that a panic is appropriate in this case (it's like accessing an array out of bounds).

server/src/metrics.rs
honzasp: This would panic if the `gauge_opts` were invalid (e.g. if `"version"` is not a valid metric name).

server/src/metrics.rs
honzasp: Yes, the user controls the options and they might trigger a panic, it would be better if this was handled by returning an error.

server/src/metrics.rs
honzasp: The conversion to `String` is probably unnecessary, this function may simply return `Vec<u8>`

PR #1872, opened 2022-10-26
Mongo filtering

BearLemma:
This PR adds support for mongo-like filtering syntax to our TS API.
After several iterations, I've landed on the approach you see in this PR, in the end, it's pretty simplistic. I provide both SQL and TS-side filtering (the latter as fallback when compiler can't optimize filtering lambdas).

One little inconsistency I've noticed is that if you filter by an JS object `{fieldName: undefined}`, it works as a Noop if send to SQL because the filter is first serialized to JSON which results in JSON: `{}`, but my TS filtering compares the undefined value present in the object to given field. Not sure what to do about that...

Tracking issue - https://github.com/chiselstrike/chiselstrike/issues/1839


api/src/datastore.ts
```diff
@@ -324,6 +326,42 @@ class ExpressionFilter<T> extends Operator<T, T> {
     }
 }
 
+/**
+ * ExpressionFilter operator is intended only to be used by Chisel compiler.
+ * It applies `predicate` on each element and keeps only those for which
+ * the `predicate` returns true. The Chisel compiler provides an `expression`
+ * as well which is to be equivalent to the predicate and which is sent to
+ * the Rust backend for direct Database evaluation if possible.
+ */
```
honzasp, line 335: This comment does not seem to be intended for this class?

api/src/datastore.ts
```diff
@@ -540,9 +578,16 @@ export class ChiselCursor<T> {
      * object `restrictions`.
      */
     filter(restrictions: Partial<T>): ChiselCursor<T>;
+    /**
+     * Restricts this cursor to contain only such elements that match the FilterExpr
+     * object. The FilterExpr object follows the syntax of MongoDB query objects.
```
honzasp, line 583: I'm not sure whether we want to explicitly say that we are copying MongoDB: it looks weird, and sooner or later we will become incompatible.

honzasp: (This relates to all mentions of the word "Mongo" in this PR)

api/src/filter.ts
```diff
@@ -0,0 +1,76 @@
...
+            } else if (cmpOp == "$lte") {
+                return v <= filterValue;
+            } else {
+                throw Error(
+                    `trying to filter with an unexpected operator '${cmpOp}'`,
+                );
+            }
+        });
+    } else {
+        return v == filter;
```
honzasp, line 74: Is the use of `==` instead of `===` intentional? (here and elsewhere) I'm not sure which option is better, but `===` is less surprising, so we might prefer to use it here. (For example, `"100" == 100` is true)

server/src/datastore/filter.rs
```diff
@@ -0,0 +1,82 @@
...
+                for operand in operands {
+                    expr = BinaryExpr::or(expr, to_expr(operand)?)
+                }
+                expr
+            }
+            "$not" => Expr::Not(to_expr(value)?.into()),
+            field_name => {
+                anyhow::ensure!(
+                    !field_name.starts_with('$'),
+                    "found an unknown filter operator '{field_name}'"
```
honzasp, line 33: `{field_name:?}` is usually better than `'{filter_name}'`, because it ensures that even weird strings (like `"\n"` or `"'"`) are always formatted nicely.

server/src/datastore/filter.rs
```diff
@@ -0,0 +1,82 @@
...
+                );
+                filed_filter_to_expr(field_name, value)?
+            }
+        };
+        expr = BinaryExpr::and(expr, e);
+    }
+    Ok(expr)
+}
+
+fn filed_filter_to_expr(field_name: &str, value: &serde_json::Value) -> Result<Expr> {
```
honzasp, line 43: s/filed/field

api/src/filter.ts
```diff
@@ -0,0 +1,76 @@
...
+            } else if (cmpOp == "$gte") {
+                return v >= filterValue;
+            } else if (cmpOp == "$lt") {
+                return v < filterValue;
+            } else if (cmpOp == "$lte") {
+                return v <= filterValue;
+            } else {
+                throw Error(
+                    `trying to filter with an unexpected operator '${cmpOp}'`,
+                );
```
honzasp, line 70: Hmm, the Rust code seems to accept filters like `{ author: { age: {$gt: 40} } }`, but this code rejects them?

honzasp (commented):
Looks very nice! :) Just a few comments related to details.

api/src/datastore.ts
honzasp: Perhaps we should use "filter expression" in code, and say that they are "similar to Mongo queries" in the documentation?

api/src/datastore.ts
```diff
@@ -324,6 +326,42 @@ class ExpressionFilter<T> extends Operator<T, T> {
...
 }
 
+/**
+ * ExpressionFilter operator is intended only to be used by Chisel compiler.
+ * It applies `predicate` on each element and keeps only those for which
+ * the `predicate` returns true. The Chisel compiler provides an `expression`
+ * as well which is to be equivalent to the predicate and which is sent to
+ * the Rust backend for direct Database evaluation if possible.
+ */
+class MongoFilter<T> extends Operator<T, T> {
```
penberg, line 336: Let's not call this `MongoFilter`, but something else like `ChiselFilter` because we don't _aim_ at MongoDB compatibility.

api/src/datastore.ts
```diff
@@ -324,6 +326,42 @@ class ExpressionFilter<T> extends Operator<T, T> {
     }
 }
 
+/**
+ * ExpressionFilter operator is intended only to be used by Chisel compiler.
```
haaawk, line 330: s/ExpressionFilter/MongoFilter/ ? Seems like a copy paste error, no?

server/src/datastore/filter.rs
```diff
@@ -0,0 +1,84 @@
+use crate::datastore::expr::{BinaryExpr, BinaryOp, Expr, PropertyAccess, Value as ExprValue};
+use anyhow::{Context, Result};
+
+pub fn to_expr(e: &serde_json::Value) -> Result<Expr> {
```
haaawk, line 4: This function is recursive and the depth of the recursion depends on the user input. Would it make sense to put a hard limit on the depth of the recursion to make sure that we don't end up with stack overflow?

I know that in practice this should never happen for any reasonable input but it still feels wrong to allow it to happen at all.

server/src/datastore/filter.rs
```diff
@@ -0,0 +1,84 @@
...
+
+fn filed_filter_to_expr(field_name: &str, value: &serde_json::Value) -> Result<Expr> {
+    let property = PropertyAccess {
+        property: field_name.to_owned(),
+        object: Expr::Parameter { position: 0 }.into(),
+    };
+    field_filter_to_expr_rec(property, value)
+}
+
+fn field_filter_to_expr_rec(property: PropertyAccess, value: &serde_json::Value) -> Result<Expr> {
```
haaawk, line 51: Same comment about limiting recursion.

api/src/filter.ts
```diff
@@ -0,0 +1,76 @@
...
+    "$or"?: FilterExpr[];
+    "$not"?: FilterExpr;
+} | Record<string, FieldFilter>;
+
+export function evalFilter(
+    filterExpr: FilterExpr,
+    v: Record<string, unknown>,
+): boolean {
+    return Object.entries(filterExpr).every(([key, filter]) => {
+        if (key === "$and") {
```
haaawk, line 26: Would it make sense to use switch?

api/src/filter.ts
```diff
@@ -0,0 +1,76 @@
...
+            if (v === null || v === undefined) {
+                throw Error(
+                    `can't apply comparison operator ${cmpOp} to a value '${v}'`,
+                );
+            } else if (filterValue === null || filterValue === undefined) {
+                throw Error(
+                    `can't apply comparison operator ${cmpOp} to a filter value '${filterValue}'`,
+                );
+            }
+            if (cmpOp === "$gt") {
```
haaawk, line 59: switch?

api/src/filter.ts
```diff
@@ -0,0 +1,106 @@
...
+        return Object.entries(filter).every(([key, filterValue]) => {
+            if (valueIsObj) {
+                return evalFieldFilter(filterValue, v[key]);
+            } else {
+                return evalOperator(
+                    key as ComparisonOperator,
+                    filterValue as PrimitiveValue,
+                    v as PrimitiveValue,
+                );
+            }
```
honzasp, line 62: If I understand this correctly, this means that if filter is `{$gt: 40}` and the value is `30`, we will evaluate `30 > 40`, but if the filter is `{$gt: 40}` and the value is `{age: 30}`, we will try to evaluate filter `40` against value `undefined` (`{age: 30}["$gt"]`). Is that expected behavior?

api/src/filter.ts
```diff
@@ -0,0 +1,106 @@
...
+        });
+    } else {
+        if (valueIsObj) {
+            throw Error(
+                `failed to filter with filter value '${filter}' against field value '${v}'`,
+            );
```
honzasp, line 68: I'm not sure whether it's a good idea to include a string representation of `v` in the error. First, this is JavaScript, so the string representation may be confusing (such as `""` for `[]`, `"1"` for `[1]`, `"\n"` for `"\n"` or `"[object Object]"` for `{age: 30}`). Second, the value may be very large or it may contain sensitive data, so it might not always be appropriate to include it in the error message.

api/src/filter.ts
```diff
@@ -0,0 +1,106 @@
...
+    v: PrimitiveValue,
+) {
+    if (cmpOp === "$eq") {
+        return v === filterValue;
+    } else if (cmpOp === "$ne") {
+        return v !== filterValue;
+    }
+    if (v === null || v === undefined) {
+        throw Error(
+            `can't apply comparison operator ${cmpOp} to a value '${v}'`,
```
honzasp, line 86: It might be a good idea to sanitize the `cmpOp` using `JSON.stringify(cmpOp)` (in the same way that it's a good practice to print unknown strings for debugging with `{:?}` in Rust)

haaawk (approved):

PR #1875, opened 2022-10-27
tests: Create Kafka topic in test

penberg:
This paves way for having multiple test cases that use their own topics
for test isolation.

Please note that the created topics are not deleted after test run
because rskafka does not support it:

https://github.com/influxdata/rskafka/issues/190

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -178,7 +178,6 @@ jobs:
           tar xf kafka_2.12-3.2.3.tgz
           ./kafka_2.12-3.2.3/bin/zookeeper-server-start.sh -daemon kafka_2.12-3.2.3/config/zookeeper.properties
           ./kafka_2.12-3.2.3/bin/kafka-server-start.sh -daemon kafka_2.12-3.2.3/config/server.properties
-          ./kafka_2.12-3.2.3/bin/kafka-topics.sh --create --topic testing --bootstrap-server localhost:9092
 
     - name: Run Kafka integration tests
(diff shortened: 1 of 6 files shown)


honzasp:
Hmm, but this still uses a hard-coded topic for each test, right? Would it make sense if the test harness itself created (and then removed) the topic? Something like

```rust
#[self::test(modules = Node, kafka_topics = 1)]
async fn test_kafka_consume(c: TestContext) {
    // ...
    let kafka_topic = c.kafka_topic(0);
    // ...
}
```

Where the `kafka_topics` field specifies the number of topics that are to be created (defaults to 0), and the `TestContext::kafka_topic(i: usize) -> String` method returns the name of topic with index `i`?

One way to create the topics would be to generate large random names (e.g. uuids).

honzasp:
> Where is the right place to delete the topics?

The deletion of topics will probably be `async`, right? This cannot be done in the destructor of `TestContext`, so perhaps the best option would be to add a `cleanup()` method to `TestContext` and call it from `run_tests()` in `cli/tests/integration_tests/rust.rs` (after the `instance.spec.test_fn.call(ctx).await;` call). However, this will require changing the tests to take a `&mut TestContext` instead of `TestContext`.

penberg:
@honzasp Unfortunately the `rskafka` create does *not* support deletion of topics at this time so will have to defer that for later.

honzasp (approved):
Nice solution! :) It does not provide isolation between different runs of the same test suite (both concurrent and sequential), but it should be enough for now.

PR #1876, opened 2022-10-27
Policies: pass SourceMap to emit js

MarinPostma:
This fixes a bug where swc would panic when emiting code, because it is not passed the `SourceMap` that was used for parsing. This is fixed by passing to the emmiter the same `SourceMap` that was used for parsing.


MarinPostma:
I tried to reproduce it, and I don't know how to actually test that. I hit this issue when emitting bad code-gen, so if that happens, I would hit other tests as well.

PR #1879, opened 2022-10-27
Introduce Nursery

honzasp:
Introduce the `Nursery` to abstract away the common pattern of awaiting a dynamic set of futures. This unifies the manual polling in `Trunk` and the suboptimal code in `KafkaService`.

### DIFF
--- a/api/src/kafka.ts
+++ b/api/src/kafka.ts
@@ -16,9 +16,8 @@ export class TopicMap {
         this.topics = {};
     }
 
-    async topic(topic: string, handler: EventHandler) {
+    topic(topic: string, handler: EventHandler) {
         this.topics[topic] = handler;
-        await opAsync("op_chisel_subscribe_topic", topic);
     }
 }
 
--- a/api/src/run.ts
+++ b/api/src/run.ts
@@ -33,10 +33,14 @@ export default async function run(
     specialAfter(routeMap);
     const router = new Router(routeMap);
 
+    // subscribe to all requested Kafka topics
     const topicMap = userTopicMap ?? new TopicMap();
+    for (const topic in topicMap.topics) {
+        opSync("op_chisel_subscribe_topic", topic);
+    }
 
     // signal to Rust that we are ready to handle jobs
-    Deno.core.opSync("op_chisel_ready");
+    opSync("op_chisel_ready");
 
     // register new error class
     // @ts-ignore: Dynamic property
--- a/server/src/kafka.rs
+++ b/server/src/kafka.rs
@@ -1,13 +1,12 @@
 // SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
 
-use crate::internal::is_stopping;
+use crate::nursery::{Nursery, NurseryStream};
 use crate::server::Server;
 use crate::version::VersionJob;
-use anyhow::{anyhow, Result};
+use anyhow::{Result, Context};
 use deno_core::serde_v8;
 use enclose::enclose;
-use futures::stream::{AbortHandle, Abortable, FuturesUnordered, StreamExt, TryStreamExt};
-use futures::FutureExt;
+use futures::stream::{StreamExt, TryStreamExt, FuturesUnordered};
 use parking_lot::Mutex;
 use rskafka::client::{
     consumer::{StartOffset, StreamConsumerBuilder},
@@ -18,7 +17,6 @@ use rskafka::record::Record;
 use serde::Serialize;
 use std::collections::HashMap;
 use std::sync::Arc;
-use tokio::sync::Notify;
 use utils::TaskHandle;
 
 /// Kafka event that is passed to JavaScript.
@@ -33,7 +31,8 @@ pub struct KafkaEvent {
 pub struct KafkaService {
     client: Client,
     topics: Mutex<HashMap<String, Arc<PartitionClient>>>,
-    resubscribe: Notify,
+    topic_nursery: Nursery<TaskHandle<Result<()>>>,
+    topic_stream: Mutex<Option<NurseryStream<TaskHandle<Result<()>>>>>,
 }
 
 impl KafkaService {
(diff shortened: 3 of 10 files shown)


honzasp:
Indeed! Neither I nor the author of `async_nursery` came up with this:
https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/

I wanted to use `async_nursery` about a year ago, but I didn't like it, so I implemented my own version, which is much simpler. It is also more general, because it works with any futures, not just task handles.

MarinPostma:
ah yes that's why this rang a bell :)

Since this is so general, why not make it into a crate?

api/src/kafka.ts
```diff
@@ -16,9 +16,8 @@ export class TopicMap {
         this.topics = {};
     }
 
-    async topic(topic: string, handler: EventHandler) {
+    topic(topic: string, handler: EventHandler) {
```
MarinPostma, line 19: this method name is not very descriptive, it really sounds like a getter

MarinPostma: maybe something like `subscribe` would be more appropriate?

api/src/run.ts
```diff
@@ -33,10 +33,14 @@ export default async function run(
     specialAfter(routeMap);
     const router = new Router(routeMap);
 
+    // subscribe to all requested Kafka topics
     const topicMap = userTopicMap ?? new TopicMap();
+    for (const topic in topicMap.topics) {
+        opSync("op_chisel_subscribe_topic", topic);
+    }
```
MarinPostma, line 40: maybe this op should take an array of topics then?

server/src/kafka.rs
```diff
@@ -42,85 +41,51 @@ impl KafkaService {
...
+) -> Result<()> {
+    let mut stream = StreamConsumerBuilder::new(client, StartOffset::Latest)
+        .with_max_wait_ms(100)
+        .build();
+    while let Some(event) = stream.next().await {
+        let (record_and_offset, _) =
+            event.context("Could not receive Kafka event from consumer")?;
+        handle_event(&server, topic.clone(), record_and_offset.record).await?;
+    }
+    Ok(())
```
MarinPostma, line 88: @penberg wasn't this code supposed to stall on resubscription?

server/src/nursery.rs
```diff
@@ -0,0 +1,99 @@
...
+where
+    Fut: Future<Output = T>,
+{
+    type Item = T;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<T>> {
+        let this = self.get_mut();
+        guard! {let Some(state) = Weak::upgrade(&this.state) else {
+            return Poll::Ready(None);
+        }};
```
MarinPostma, line 81: next rust version we won't need this macro anymore :)

honzasp: Yeah, I noticed that there was some progress :) I'm looking forward to it, it's my favorite piece of non-standard syntax!

api/src/kafka.ts
honzasp: The `TopicMap` was intended to be an analogy to `RouteMap`, so `TopicMap` uses `topic()` to define a topic, in the same way that `RouteMap` uses `route()` to define a route.

(Note that this API is only used from generated code at the moment.)

api/src/run.ts
honzasp: Hmm, when we need to do a thing multiple times, it seems preferable to provide a simple primitive op that does the thing and call it in a loop, rather than provide a "vectorized" primitive op that receives an array and does the thing multiple times.

server/src/kafka.rs
honzasp: I'm not sure I understand: with this code, there are no resubscriptions. Once the server subscribes to a topic, it stays subscribed until it terminates.

honzasp:
> Since this is so general, why not make it into a crate?

I did consider putting this into `utils`, together with `TaskHandle` and friends, but I decided to put it into `server` first, and move it into `utils` only if we find that it's actually useful outside of the server.

api/src/run.ts
MarinPostma: Gut feeling?

server/src/kafka.rs
MarinPostma: Oh ok, so the main difference is that each topic is handled independently in a task?

server/src/nursery.rs
MarinPostma: Is officially apart of the release for next week ☺️

MarinPostma: Along with GAT

server/src/kafka.rs
honzasp: Yes, the task is spawned when the user code first subscribes for it, and we never unsubscribe.

api/src/run.ts
honzasp: Not just gut feeling, I think that it is a standard practice to provide an interface that does a thing, rather than providing an interface that does N things sequentially, unless there is a good reason (usually performance). For example, most POSIX syscalls (which I tend to view as a good analogy to ops), such as `write()`, `fsync()`, `close()`, `pipe()` or `socket()`, work on just a single descriptor, and syscalls that are just a vectorized versions (such as `close_range()` or `writev()`) are provided only when it really makes sense.

---

If you want some abstract reasoning, I think it's usually preferable to provide smaller pieces that compose well together (in this case, the pieces are the for-loop and the op), rather than pre-composed blocks (a hypothetical op that subscribed to multiple topics at once). This allows you to have smaller number of pieces and greater freedom in composing them.

Think of the difference between Lego and Barbies

MarinPostma:
looks good to me, @penberg I'll let you take a look and approve since this is also changing some of your code

MarinPostma:
I don't think Pekka will have time to review that. LGTM anyways, let's merge.

MarinPostma (approved):

PR #1882, opened 2022-11-01
Implement secrets polling period flag.

BearLemma:


server/src/opt.rs
```diff
@@ -63,6 +63,10 @@ pub struct Opt {
     #[structopt(long, env = "CHISEL_SECRET_LOCATION")]
     pub chisel_secret_location: Option<String>,
 
+    /// Sets secrets polling period in seconds (can be float).
+    #[structopt(long, default_value = "1")]
```
haaawk, line 67: Maybe make this millis in case we ever want to poll more often than a second?

honzasp: In that case, perhaps this should be a float?

server/src/server.rs
```diff
@@ -278,7 +278,10 @@ async fn refresh_secrets(server: Arc<Server>) -> Result<()> {
         } else {
             last_try_was_failure = false;
         }
-        tokio::time::sleep(Duration::from_millis(1000)).await;
+        tokio::time::sleep(Duration::from_millis(
+            (1000.0 * server.opt.secrets_polling_period_s) as u64,
+        ))
+        .await;
```
honzasp, line 284: I would use `Duration::from_secs_f32(server.opt.secrets_polling_period_s)`

haaawk (approved):

haaawk:
Build fixed here https://github.com/chiselstrike/chiselstrike/pull/1883

PR #1884, opened 2022-11-01
Flag to disable secrets polling

athoscouto:


haaawk:
LGTM. I think it can solve our problem with CPU utilisation. The only thing that I would like to see added to the PR is that we stop refreshing only after we successfully obtained secrets for the first time to prevent a situation when we have a deployment without secrets.

server/src/opt.rs
```diff
@@ -67,6 +67,10 @@ pub struct Opt {
     #[structopt(long, default_value = "1")]
     pub secrets_polling_period_s: f32,
 
+    /// Disables periodic polling to update secrets.
```
haaawk, line 70: Proposal: /// Disables periodic polling to update secrets after secrets are successfully obtained for the first time.

haaawk: Did you update the comment?

server/src/server.rs
```diff
@@ -267,7 +267,7 @@ async fn start_builtin_version(server: Arc<Server>) -> Result<()> {
     Ok(())
 }
 
-async fn refresh_secrets(server: Arc<Server>) -> Result<()> {
+async fn refresh_secrets(server: Arc<Server>, stop_on_success: bool) -> Result<()> {
```
honzasp, line 270: It's not necessary to pass the extra argument, you can read it directly from `server.opt`.

server/src/server.rs
```diff
@@ -277,6 +277,10 @@ async fn refresh_secrets(server: Arc<Server>) -> Result<()> {
             last_try_was_failure = true;
         } else {
             last_try_was_failure = false;
+
+            if stop_on_success {
+                return Ok(());
+            }
```
honzasp, line 283: This mechanism will typically load the secrets twice: once in `make_server()`, and then here in `refresh_secrets()`. Maybe it would be better to enforce that `make_server()` always loads valid secrets (by returning the error from `get_secrets()` instead of just logging it and using empty secrets), and never start the `refresh_secrets()` task if `opt.disable_secrets_polling` is true?

haaawk: I think it's fine for code simplicity and change small size.

haaawk:
```
failures:

---- linters::check_formating stdout ----

thread 'linters::check_formating' panicked at 'Command "cargo" ["fmt", "--", "--check"] exited with status exit status: 1', cli/tests/common/mod.rs:34:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    linters::check_formating
```

server/src/server.rs
haaawk: That was my initial thought too but the I reflected that naming the parameter and decoupling it from the flag makes sense too.

server/src/server.rs
honzasp: I think that making sure that we always have valid secrets and not spawning a task to refresh secrets if we don't want to refresh secrets is even simpler :)

haaawk: What would happen if we start creating launchpad project and it fails to get secrets?

haaawk:
Here's a version that should pass the build check and that adds exponential backoff -> https://github.com/chiselstrike/chiselstrike/pull/1885

server/src/server.rs
athoscouto: Honza's proposal do not mention it explicitly. But, IMO, it makes sense to skip loading secrets on apply - so that won't happen.

server/src/server.rs
```diff
@@ -138,9 +138,13 @@ async fn make_server(opt: Opt) -> Result<(Arc<Server>, TaskHandle<Result<()>>)>
...
 
-    let secrets = match secrets::get_secrets(&opt).await {
-        Ok(secrets) => secrets,
-        Err(err) => {
+    let secrets = match (
+        secrets::get_secrets(&opt).await,
+        opt.disable_secrets_polling,
+    ) {
+        (Ok(secrets), _) => secrets,
+        (Err(err), true) => return Err(err),
```
haaawk, line 146: I don't think it's a good idea to fail startup just because secrets can't be read. We even have tests that check the startup is successful even when secrets can't be read.

haaawk:
Already merged here https://github.com/chiselstrike/chiselstrike/pull/1885

server/src/server.rs
honzasp: I think that if the server fails to load secrets at startup, it should terminate with an error (and probably get restarted by the supervisor). If the user code expects to have defined secrets, but we behave like no secrets are given, bad things may happen.

(However, I'm getting out of scope of this PR :) )

PR #1894, opened 2022-11-03
Add .routegen to gitignore template

CodingDoug:


packages/create-chiselstrike-app/template/gitignore
```diff
@@ -2,3 +2,4 @@
 /.gen
 /.vscode
 /node_modules
+/.routemap
```
dekimir, line 5: Isn't it `.routegen`??

dekimir (changes requested):

packages/create-chiselstrike-app/template/gitignore
honzasp: Indeed: https://github.com/chiselstrike/chiselstrike/blob/be49dbe32a7ff303d3db0899b4864dd1e99aad00/cli/src/cmd/apply/node.rs#L95

dekimir (approved):

PR #1895, opened 2022-11-04
server/kafka: Don't kill server on receive failure

penberg:
It's easy to trigger the receive failure path. For example, add a topic handler to the `events` directory for a non-existing topic. So let's not kill the server when that happens, but instead, warn about it, which is better behavior for transient falures as well.

Spotted by the `test_kafka_apply` test case that -- for some reason -- started failing on my machine.

### DIFF
--- a/server/src/kafka.rs
+++ b/server/src/kafka.rs
@@ -3,7 +3,7 @@
 use crate::nursery::{Nursery, NurseryStream};
 use crate::server::Server;
 use crate::version::VersionJob;
-use anyhow::{Context, Result};
+use anyhow::Result;
 use deno_core::serde_v8;
 use enclose::enclose;
 use futures::stream::{FuturesUnordered, StreamExt, TryStreamExt};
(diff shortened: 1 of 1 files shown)


server/src/kafka.rs
```diff
@@ -81,9 +81,14 @@ async fn handle_topic(
...
-        let (record_and_offset, _) =
-            event.context("Could not receive Kafka event from consumer")?;
-        handle_event(&server, topic.clone(), record_and_offset.record).await?;
+        match event {
+            Ok((record_and_offset, _)) => {
+                handle_event(&server, topic.clone(), record_and_offset.record).await?;
+            }
+            Err(err) => {
+                warn!("Failed to receive Kafka event: {}", err);
+            }
```
honzasp, line 90: Are we sure that if an error happens, it's returned from `.next()` just once? If the stream enters a state where `.next()` always return an error, this will fall into a nasty infinite loop.

honzasp: Possible solutions for this are:
1. Include a delay after every error (on the order of 1 second)
2. Limit the number of errors that we tolerate; if we exceed this number, return the error to the caller (which should crash the server in the end).

penberg: The contract between rskafka and us is that _if_ the stream becomes faulty, it should terminate and `next()` returns `None`, which causes us to exit the loop.

But as long as the stream is operational, it's totally fine to keep hammering the loop. Yes, we could be spamming the log very hard, but that's a general problem that hopefully should be addressed at the logger level.

And btw, the documentation suggests that if the stream consumer fails to fetch records, the error is only emitted once:

> If an error is returned by `fetch_records` then the stream will emit this error once and will terminate afterwards.

(Source: https://docs.rs/rskafka/latest/rskafka/client/consumer/struct.StreamConsumer.html#error-handling)

So @honzasp I don't think any code change is necessary.

honzasp (approved):

PR #1908, opened 2022-11-22
Add support for FilterExpr in find_many and find_one.

BearLemma:
As requested by @CodingDoug , this PR brings support for FilterExpr to both findMany and findOne functions in the TS API. I've also made the filterExpr a little bit stricter so that it doesn't accept filters for keys which are not part of the entity.


CodingDoug:
`ChiselEntity.delete()` and `ChiselEntity.upsert()` also do filtering with restrictions (but not predicate). How about adding FilterExpr to those?

CodingDoug (approved):

PR #1918, opened 2022-12-02
Generate clients

BearLemma:
This PR brings support for client generation.
The main changes include

- New build container to include `ts-node`. 
- `chisel generate` generates TS client that can be used against our CRUD endpoints
- `/__chiselstrike/routes` internal reflection endpoint returning information about routing
- Added support for testing of generated clients to our rust integration tests.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -36,7 +36,7 @@ jobs:
 
     - uses: actions/setup-node@v3
       with:
-          node-version: 16
+          node-version: 18
 
     - name: Install nightly rust toolchain
       uses: actions-rs/toolchain@v1
@@ -54,6 +54,9 @@ jobs:
       with:
             path: ~/.npm
             key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+
+    - name: Install ts-node
+      run: npm install -g ts-node
 
     - name: Cargo config
       run: |
@@ -116,13 +119,16 @@ jobs:
     - name: Install node
       uses: actions/setup-node@v3
       with:
-        node-version: 16
+        node-version: 18
 
     - name: Configure Node cache
       uses: actions/cache@v2
       with:
             path: ~/.npm
             key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
+
+    - name: Install ts-node
+      run: npm install -g ts-node
 
     - name: Cargo config
       run: |
--- a/cli/src/cmd/generate.rs
+++ b/cli/src/cmd/generate.rs
@@ -1,53 +1,452 @@
 use crate::proto::chisel_rpc_client::ChiselRpcClient;
 use crate::proto::{type_msg::TypeEnum, DescribeRequest};
-use anyhow::{anyhow, Result};
+use crate::proto::{FieldDefinition, TypeDefinition, TypeMsg, VersionDefinition};
+
+use anyhow::{anyhow, Context, Result};
+use serde::Deserialize;
+use serde_json::json;
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::Write;
 use std::path::PathBuf;
+use tokio::fs::create_dir_all;
 
-pub(crate) async fn cmd_generate(
-    server_url: String,
-    output: PathBuf,
-    version: String,
-) -> Result<()> {
-    let output = File::create(output)?;
-    let mut client = ChiselRpcClient::connect(server_url).await?;
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Mode {
+    Node,
+    Deno,
+}
+
+#[derive(Debug)]
+pub struct Opts {
+    pub server_url: String,
+    pub api_addres: String,
+    pub output_dir: PathBuf,
+    pub version: String,
+    pub mode: Mode,
+}
+
+pub(crate) async fn cmd_generate(opts: Opts) -> Result<()> {
+    create_dir_all(&opts.output_dir)
+        .await
+        .context("failed to create directory for generated client files")?;
+
+    let mut client = ChiselRpcClient::connect(opts.server_url.to_owned()).await?;
     let request = tonic::Request::new(DescribeRequest {});
     let response = execute!(client.describe(request).await);
     let version_def = response
         .version_defs
         .iter()
-        .find(|def| def.version_id == version);
-    if let Some(version_def) = version_def {
-        for def in &version_def.type_defs {
-            writeln!(&output, "export type {} = {{", def.name)?;
-            writeln!(&output, "    id: string;")?;
-            for field in &def.field_defs {
-                let field_type = field.field_type()?;
-                writeln!(
-                    &output,
-                    "    {}{}: {}{};",
-                    field.name,
-                    if field.is_optional { "?" } else { "" },
-                    field_type,
-                    field
-                        .default_value
-                        .as_ref()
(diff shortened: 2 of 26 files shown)


cli/tests/integration_tests/rust_tests/routing.rs
```diff
@@ -587,3 +589,54 @@ async fn errors(c: TestContext) {
...
+    }
+    #[derive(Debug, Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    struct Route {
+        methods: Vec<String>,
+        path_pattern: String,
+        client_metadata: Option<CrudMeta>,
+    }
+
+    let routes_json = c.chisel.get_json("/dev/__chisel_internal/routes").await;
```
penberg, line 631: It's hard to reverse engineer from this diff what these reflection endpoints look like so few words about the design in the commit message, in-tree documentation, or source comment would be great.

Why are they _internal_ endpoints, btw? In any case, let's put them under `__chiselstrike` namespace like the non-versioned ones already are. If you want to emphasise they're internal, then `/<version>/__chiselstrike/internal` maybe?

cli/tests/integration_tests/rust.rs
```diff
@@ -225,6 +211,15 @@ async fn create_kafka_topic(connection: &str, topic: &str) {
     }
 }
 
+fn ensure_dependencies() {
+    // Find our deno version and install that. We don't use --path
+    // because that always reinstall the binary.
+    let version = get_deno_version();
+    cargo_install(&version, "deno", "deno");
+
+    run("npm", ["install", "ts-node"]);
```
penberg, line 220: I suggested putting `ts-node` in the Docker image of CI, but this is also fine.

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -0,0 +1,372 @@
...
+    if (!resp.ok) {
+        // TODO: Improve error handling
+        throw Error(
+            `failed to post an entity. Got error code ${resp.status} (${resp.statusText}) with message: '${await resp
+                .text}'`,
+        );
+    }
+}
+
+async function send_json(
```
penberg, line 27: This is generated code so not really that important, but... How did you manage to sneak `send_json` pass our ESLint rules? The function should obviously be called `sendJson` 😝

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -0,0 +1,372 @@
+export function urlJoin(...urlParts: string[]): URL {
```
penberg, line 1: We can start with this, but I'd assume much of the generated code could be a proper `npm` package such as `@chiselstrike/client` or something and the generated code could just pull that as a dependency, no?

honzasp (commented):
Nice work! There are some comments. The biggest problem is that we are adding this client code generation as an afterthought, so there are some holes which will not be easy to fix (for example, we allow path patterns like `/book{s}/(\d+)`, but how should we handle them in the client?).

cli/src/cmd/generate.rs
```diff
@@ -1,53 +1,499 @@
...
+    Options,
+    Get,
+    Post,
+    Put,
+    Delete,
+    Head,
+    Trace,
+    Connect,
+    Patch,
+}
```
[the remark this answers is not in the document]
honzasp: We probably should support `*`, because it is used for legacy endpoints (i.e., files that `export default function()` instead of exporting a `RouteMap`).

cli/src/cmd/generate.rs
```diff
@@ -1,53 +1,499 @@
...
+        }
+    };
+    write!(output, "{}", &imports)?;
+    write!(output, "{}", include_str!("generate_src/client.ts"))?;
+    let root = build_routing(routes);
+
+    writeln!(
+        output,
+        r#"
+            function ΩcreateClient(ΩserverUrl: string, ΩuserVersion?: string) {{
```
honzasp, line 466: This generated function should perhaps return a typed value.

cli/src/cmd/generate.rs
```diff
@@ -1,53 +1,499 @@
...
+}
+
+fn build_routing(routes: &Vec<RouteInfo>) -> SubRoute {
+    let mut root = SubRoute::new(RouteFragment::Static("".to_owned()));
+
+    for route in routes {
+        if route.methods != vec![HttpMethod::Post] {
+            continue;
+        }
+        // Hacky way to detect CRUD endpoints
```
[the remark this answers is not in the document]
honzasp: One way that this can break is if the user disables the `POST` route in CRUD, for example by using `{write: false}`.

.github/workflows/rust.yml
```diff
@@ -116,14 +119,17 @@ jobs:
     - name: Install node
       uses: actions/setup-node@v3
       with:
-        node-version: 16
+        node-version: 18
 
     - name: Configure Node cache
       uses: actions/cache@v2
```
barbieri, line 125: you can avoid this explicit cache with https://github.com/actions/setup-node#caching-global-packages-data + `cache: npm`

cli/src/cmd/generate_src/client_lib.ts
barbieri: it would be unique if multiple users, also usually comes built so no need to build them locally... and reduces the amount of generated code (since it's basically static)

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -0,0 +1,387 @@
...
+
+        const ensureType = (typeName: string) => {
+            if (typeof fieldValue !== typeName) {
+                throw Error(
+                    `field ${field.name} of entity ${entityName} is ${typeName}, but provided value is of type ${typeof fieldValue}`,
+                );
+            }
+        };
+        const fieldType = field.type.name;
+        if (fieldType == "string" || fieldType == "entityId") {
```
barbieri, line 73: use `===` since JS's `==` is stupid: https://eslint.org/docs/latest/rules/eqeqeq

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -0,0 +1,387 @@
...
+        );
+    } else {
+        throw new Error(
+            `field ${fieldName} of entity ${entityName} is Entity, but received value is not an object`,
+        );
+    }
+}
+
+function assertNever(x: never): never {
+    return x;
```
barbieri, line 227: maybe throw?

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -0,0 +1,387 @@
...
+                entityValue[fieldName] = undefined;
+                continue;
+            } else {
+                throw new Error(
+                    `field ${fieldName} of entity ${entityName} is not optional but undefined/null was received for the field`,
+                );
+            }
+        }
+
+        const ensureType = (typeName: string) => {
```
barbieri, line 65: `string` is too broad and typos will let mistakes pass... you should make it an union of supported literals such as `"number" | "boolean" | "string" | "object" | "function"`

barbieri (commented):
I did an initial review focused only on the TypeScript/JS stuff, looks good. I didn't look at any of Rust files.

penberg (approved):

cli/src/cmd/generate.rs
honzasp: Oh, that's great, I didn't know that TypeScript will infer the return type if you omit it!

PR #1922, opened 2022-12-06
feat(api): expose EventHandler, ResponseLike and ChiselEntityClass

ricardodalarme:
## Description

To import these files you needed to import from the file. Example:

```ts
import type { EventHandler } from '@chiselstrike/api/lib/kafka.ts';
```

Now allow to import it from `@/chiselstrike/api`

### DIFF
--- a/api/src/api.ts
+++ b/api/src/api.ts
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
 
 export { crud } from "./crud.ts";
+export type { ChiselEntityClass } from "./crud.ts";
 export {
     AuthUser,
     ChiselCursor,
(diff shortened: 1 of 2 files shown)


penberg:
@ricardodalarme Can you give an example of why you need these?

ricardodalarme:
ChiselEntityClass was a request from luis, we use it in dogfood. Maybe it won't be used in any other context than Dogfood, but also why don't we export it?

luisfvieirasilva:
@honzasp, as @ricardodalarme said, I think that would be nice to have an type that I can use to pass a ChiselEntity class (instead of not an instance).

That said, I just remembered that I wasn't able to achieve this using the original `ChiselEntityClass`. I needed to adapt it as you can see [here](https://github.com/chiselstrike/managed-services-backend/blob/main/src/utils/chiselstrikeApiExtension.ts#L17)

honzasp:
I see, you are right, I change my opinion, lets export `ChiselEntityClass` :) But it might be better to define it next to `ChiselEntity`, not in `crud.ts` (because it is no longer a CRUD implementation detail).

penberg:
@honzasp Let's approve this and you send a follow-up to do the code movement you requested?

honzasp (approved):

PR #1930, opened 2022-12-09
Improve test coverage of clients and fix found bugs

BearLemma:
In this PR, I have done the following:

- Enabled type checking for client tests and fixed the corresponding bugs that were found.
- Implemented and activated entityToJson() to perform the inverse operation of jsonToEntity().
- Introduced and used OmitRecursively for the post() function to allow the submission of entities and nested entities without an ID.
- Added tests covering the .post() function with all types that we provide (with the exception of ArrayBuffer, which will require a separate PR).

### DIFF
--- a/cli/src/cmd/generate_src/client_lib.ts
+++ b/cli/src/cmd/generate_src/client_lib.ts
@@ -350,7 +350,7 @@ export function makeGetAll<Entity>(
     origUrl: URL,
     serverUrl: string,
     entityType: reflect.Entity,
-): (params?: GetParams<Entity>) => Promise<Entity[]> {
+): (params?: GetAllParams<Entity>) => Promise<Entity[]> {
     const makeIter = makeGetManyIter<Entity>(origUrl, serverUrl, entityType);
     return async function (params?: GetAllParams<Entity>): Promise<Entity[]> {
         let iterParams = {};
--- a/cli/src/cmd/generate_src/client_lib.ts
+++ b/cli/src/cmd/generate_src/client_lib.ts
@@ -42,18 +42,42 @@ async function sendJson(
     return resp;
 }
 
+class AccessContext {
+    constructor(private context: string) {}
+
+    onField(field: string): AccessContext {
+        return new AccessContext(this.context + `.${field}`);
+    }
+
+    onArray(idx?: number): AccessContext {
+        if (idx !== undefined) {
+            return new AccessContext(this.context + `[${idx}]`);
+        } else {
+            return new AccessContext(this.context + `[]`);
+        }
+    }
+
+    toString(): string {
+        return this.context;
+    }
+}
+
 function entityFromJson<Entity>(
     entityType: reflect.Entity,
     inputValue: Record<string, unknown>,
+    entityContext?: AccessContext,
 ): Entity {
     const entityValue: Record<string, unknown> = {};
     const entityName = entityType.name;
+    entityContext = entityContext ?? new AccessContext(entityName);
+
     for (const field of entityType.fields) {
         if (!(field.name in inputValue)) {
             continue;
         }
         const fieldName = field.name;
         const fieldValue = inputValue[fieldName];
+        const context = entityContext.onField(fieldName);
 
         if (fieldValue === null || fieldValue === undefined) {
             if (field.isOptional) {
@@ -61,14 +85,14 @@ function entityFromJson<Entity>(
                 continue;
             } else {
                 throw new Error(
-                    `field '${fieldName}' of entity '${entityName}' is not optional but undefined/null was received for the field`,
+                    `${context} is not optional but undefined/null was received for the field`,
                 );
             }
         }
 
         const err = (typeName: string) => {
             return Error(
-                `field '${field.name}' of entity '${entityName}' is ${typeName}, but provided value is of type ${typeof fieldValue}`,
+                `${context} is of type ${typeName}, but provided value is of type ${typeof fieldValue}`,
             );
         };
         const fieldType = field.type.name;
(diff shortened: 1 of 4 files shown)


cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -222,13 +226,176 @@ function convertEntity(
...
+    }
+}
+
+function jsonFromEntity(
+    entityName: string,
+    fieldName: string,
+    entityType: reflect.Entity,
+    entityValue: unknown,
+    allowMissingId: boolean,
+): Record<string, unknown> {
```
honzasp, line 387: It's a bit strange that all functions that convert to JSON are named `*ToJson()`, but this function is named `jsonFromEntity()`. Also, the naming of the parameters is confusing, because some of them refer to the "parent" entity, while others refer to the "child" entity (i.e., `entityName` and `entityType` refer to different entities).

What would you think about something like this?

```javascript
function nestedEntityToJson(
    parentName: string,
    parentField: string,
    nestedType: reflect.Entity,
    nestedValue: unknown,
    allowMissingId: boolean,
)
```

honzasp: If you decide to rename the parameters, it might also be helpful to rename the first two parameters of `dateToJson` and `arrayToJson`, too, so that these functions remain consistent with each other.

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -222,13 +226,176 @@ function convertEntity(
...
+): unknown[] {
+    if (Array.isArray(arrayValue)) {
+        const elementTypeName = elementType.name;
+        switch (elementTypeName) {
+            case "string":
+            case "number":
+            case "boolean":
+            case "entityId":
+            case "arrayBuffer":
+                return arrayValue;
```
honzasp, line 343: Would it make sense to check that the elements of `arrayValue` have correct types? Also, I think that it will be necessary to convert `ArrayBuffer`-s to base64.

Perhaps it would be possible to unify the `switch` logic in `entityToJson` and `arrayToJson`?

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -222,13 +226,176 @@ function convertEntity(
...
+                        elementType.elementType,
+                        e,
+                        allowMissingId,
+                    )
+                );
+            case "entity":
+                return arrayValue.map((e) => {
+                    jsonFromEntity(
+                        entityName,
+                        fieldName,
```
honzasp, line 362: This might create confusing error messages. For example, if I have an entity like this:

```
class Party {
    dates: Array<Date>;
}
```

and I try to convert the object `{ dates: [42] }` to JSON, I will get an error message that says "field 'dates' of entity 'Party' is Date, but provided value is of different type", even though the field `dates` of entity `Party` is `Array<Date>`, not `Date`.

Maybe it would be better to pass a generic "context" (instead of `(entityName, fieldName)` pair) to the `*ToJson()` functions, so that I can get an error message that says "`Party.dates[0]` has type `Date`, but provided value is of a different type"?

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -379,13 +546,27 @@ export function makeGetAll<Entity>(
...
 }
 
-export function makePostOne<Entity>(
+// This magic is necessary to allow passing of nested objects without ID. However, once we allow plain objects
+// we will have to generate the ID-less entities explictly because otherwise we might accidentaly
+// remove 'id' fields from plain objects.
+type OmitDistributive<T, K extends PropertyKey> = T extends
+    Record<string, unknown> ? OmitId<OmitRecursively<T, K>> : T;
+// deno-lint-ignore ban-types
+type OmitId<T> = {} & { [P in keyof T]: T[P] }; // This forces the compiler to explicitly show the recursive objects
```
honzasp: The name `OmitId` is a bit confusing, because it does not omit the `id` property, but it forces the typechecker to expand `T`. Maybe `OmitExpand` or something like this would be better?

honzasp (commented):
Looks nice, just a few comments.

cli/src/cmd/generate_src/client_lib.ts
honzasp: In that case, maybe it's better to have a simpler error message that does not include the field, rather than having an error message that might be misleading?

cli/src/cmd/generate_src/client_lib.ts
honzasp: Hmm, it worked for me in the playground, but there might be some effects which I don't understand. Let's just use whatever works for you :)

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -38,33 +42,57 @@ async function sendJson(
...
+            return new AccessContext(this.context + `[${idx}]`);
+        } else {
+            return new AccessContext(this.context + `[]`);
+        }
+    }
+
+    toString(): string {
+        return this.context;
+    }
+}
```
honzasp, line 63: This is very nice :)

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -38,33 +42,57 @@ async function sendJson(
...
+
+function entityFromJson<Entity>(
     entityType: reflect.Entity,
     inputValue: Record<string, unknown>,
+    entityContext?: AccessContext,
 ): Entity {
     const entityValue: Record<string, unknown> = {};
     const entityName = entityType.name;
+    entityContext = entityContext ?? new AccessContext(entityName);
```
honzasp, line 72: It might be better to use a static method of `AccessContext`, something like `AccessContext.entity(entityName)`, instead of `new AccessContext(entityName)`, because the string inside `AccessContext` is considered private.

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -150,85 +172,250 @@ function convertArrayBuffer(
...
+                throw err("number");
+            }
+            outputJson[fieldName] = fieldValue;
+        } else if (fieldType === "boolean") {
+            if (typeof fieldValue !== "boolean") {
+                throw err("boolean");
+            }
+            outputJson[fieldName] = fieldValue;
+        } else if (fieldType === "arrayBuffer") {
+            // JSON.stringify performs automatic conversion.
```
honzasp, line 298: It seems that this comment was left over from a previous versionof the code?

honzasp (approved):
Good job, this is very nice :)

PR #1940, opened 2022-12-12
chore(third_party): upgrade to deno_std 0.167.0

ricardodalarme:
## Description

Upgrade deno_ std to v0.167 to extend support for some node libs, such as crypto

note: I just updated the submodule, I don't know if I need to do any extra configuration

## Release Notes

- https://github.com/denoland/deno_std/releases/tag/0.167.0


honzasp:
If you want to use some new Node polyfills from deno-std/node, you will also need to add new entries to `NODE_POLYFILLS` in `server/src/module_loader.rs`. You can refer to the analogous list from Deno sources:
https://github.com/denoland/deno/blob/main/cli/node/mod.rs

penberg (approved):

penberg:
@honzasp Can you please send a pull request to add those entries?

PR #1943, opened 2022-12-13
Add option to specify default headers to be used with ChiselClient.

BearLemma:


cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -1,3 +1,28 @@
+export type ClientParams = {
```
penberg, line 1: Some JSDocs/TSDocs would be useful, I think. But maybe you want to work with @CodingDoug on that?

cli/src/cmd/generate_src/client_lib.ts
```diff
@@ -1,3 +1,28 @@
...
+    headers?: Headers | Record<string, string>;
+};
+
+export type InternalClientParams = {
+    version?: string;
+    headers?: Headers;
+};
+
+export function cliParamsToInternal(
+    cliParams?: ClientParams,
```
penberg, line 12: You managed to confuse me with `cli` which 99.999% of the time means command line interface. If you're a lazy typist and can't be bothered to write 'clientParams', just call it `params` or `p` :)

penberg (approved):
lgtm, some minor stylistic comments added

cli/src/cmd/generate_src/client_lib.ts
CodingDoug: Yes, we can definitely add some TSdoc to all the exported classes and type. Let's do that in another PR.

PR #1944, opened 2022-12-13
Implement and use `valueToJson` function.

BearLemma:
This function turns any JavaScript value to JSON while converting values like ArrayBuffer and Date into chiselstrike compatible format. (e.g. ArrayBuffer is turned into base64 encoded string).

Usage of this function in `responseFromJson` fixes the issue of CRUD endpoints returning ArrayBuffer and Date in wrong format.

### DIFF
--- a/api/src/utils.ts
+++ b/api/src/utils.ts
@@ -37,7 +37,8 @@ export function responseFromJson(body: unknown, status = 200) {
     const isNullBody = status == 101 || status == 103 ||
         status == 204 || status == 205 || status == 304;
 
-    const json = isNullBody ? null : stringifyJson(body);
+    const jsonBody = valueToJson(body);
+    const json = isNullBody ? null : stringifyJson(jsonBody);
     return new Response(json, {
         status: status,
         headers: [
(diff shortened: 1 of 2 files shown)


api/src/utils.ts
```diff
@@ -73,6 +74,43 @@ function stringifyJson(value: unknown, space?: string | number): string {
...
+            binary += String.fromCharCode(bytes[i]);
+        }
+        return btoa(binary);
+    } else if (typeof v === "object") {
+        const jsonObj: { [x: string]: JSONValue } = {};
+        for (const [key, value] of Object.entries(v)) {
+            // Javascript's JSON objects omit undefined values.
+            if (value !== undefined) {
+                jsonObj[key] = valueToJson(value);
+            }
```
honzasp, line 104: Maybe we can use this opportunity to replace `undefined` with `null`, so that the keys are _not_ omitted?

api/src/utils.ts
```diff
@@ -73,6 +74,43 @@ function stringifyJson(value: unknown, space?: string | number): string {
     return JSON.stringify(value, replacer, space);
 }
 
+export function valueToJson(v: unknown): JSONValue {
+    if (v === undefined || v === null) {
+        return null;
+    } else if (typeof v === "string" || v instanceof String) {
```
honzasp, line 80: I see that you also check for `String`, would it make sense to also check for `Number` and `Boolean` in the following `if`-s?

api/src/utils.ts
```diff
@@ -73,6 +74,43 @@ function stringifyJson(value: unknown, space?: string | number): string {
...
+            if (value !== undefined) {
+                jsonObj[key] = valueToJson(value);
+            }
+        }
+        return jsonObj;
+    } else {
+        throw new Error(
+            `encountered unexpected value type '${typeof v}' when converting to JSON`,
+        );
+    }
```
honzasp, line 111: Maybe we can also check for `Map` and `Set` here, and remove the `replacer` function in `stringifyJson`?

honzasp (commented):
Looks ok, but the new `valueToJson` function should replace the `replacer` logic in `stringifyJson`.

api/src/utils.ts
honzasp: Ok, do what makes most sense to you :)

honzasp (approved):

PR #1949, opened 2022-12-19
policies: fix lookup unexisting value

MarinPostma:
This pr fixes a bug in the policies variable resolver that would return None on a lookup for an unexisting value in a header.

This is not a _very robust_ fix though. A more robust fix would differentiate between static properties and computed properties. We should aim a better compat with js.

### DIFF
--- a/chiselc/src/tools/analysis/region.rs
+++ b/chiselc/src/tools/analysis/region.rs
@@ -330,7 +330,6 @@ mod test {
             let from_node = insert_node(&mut graph, from);
             let to_node = insert_node(&mut graph, to);
 
-            dbg!((from_node, to_node, edge));
             graph.add_edge(from_node, to_node, edge);
         }
 
--- a/cli/tests/integration_tests/rust_tests/typescript_policies.rs
+++ b/cli/tests/integration_tests/rust_tests/typescript_policies.rs
@@ -265,3 +265,57 @@ async fn check_write_transform(c: TestContext) -> TestContext {
 
     c
 }
+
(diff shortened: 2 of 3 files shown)


penberg:
@MarinPostma A small test case to at least trigger the failure?

CodingDoug:
Another way to handle missing data (or anything unexpected) is to immediately deny the operation.  This would help prevent security surprises.

I believe I had also noticed that, for onRead operations, it's possible to put a null value inside an entity field (an invalid value) that came from a missing header lookup.  This sounds like bad behavior - I would have preferred that the operation simply fail with a meaningful error message in the output.

MarinPostma:
@CodingDoug that's not a good idea: it's not expected behavior, and would bite you back. Here's an example:
```typescript
(entity, context) => {
    if (context.headers["somefield"]) { // <- this would blow up!
        if (context.headers["someField"] == 42) {
            return Action.Allow;
        }
    }
}
```

CodingDoug:
As

> @CodingDoug that's not a good idea: it's not expected behavior, and would bite you back. Here's an example:
> 
> ```ts
> (entity, context) => {
>     if (context.headers["somefield"]) { // <- this would blow up!
>         if (context.headers["someField"] == 42) {
>             return Action.Allow;
>         }
>     }
> }
> ```

As it stands now with Record<string, string>, yes, that would be a problem.  But if we had a data structure contain the headers that exposes a check for existence (that doesn't ever fail), that would be fine. That way, they are specifically opting into optionality of the header with clear code.

penberg (approved):

PR #1963, opened 2022-12-21
fix convert cond to bool

MarinPostma:
This pr fixes a bug when converting from a `Cond` to a `Bool`. I was ignoring that And and Or bool can have more than 2 predicates.

I have added property testing to ensure this doesn't regress


penberg (dismissed):

chiselc/src/tools/functions.rs
```diff
@@ -23,6 +23,7 @@ impl<'a> ArrowFunction<'a> {
         match &arrow.body {
             BlockStmtOrExpr::BlockStmt(block) => {
                 let (cfg, stmt_map) = ControlFlow::build(&block.stmts)?;
+                println!("{}", cfg.dot());
```
penberg, line 26: @MarinPostma Drop this debug printout?

penberg (approved):

PR #1974, opened 2023-01-02
Drop udeps and sorted deps cargo linters to speed up test suite.

BearLemma:
As previously discussed, we can probably drop these linters as they are slow and not as important.
It's quite sufficient to run them just once in a while.


penberg (approved):
There's also another tool called `cargo-machete` which we could look into.

MarinPostma:
let's add those to the CI instead

PR #1982, opened 2023-01-10
Allow ID arrays.

BearLemma:


server/src/apply.rs
```diff
@@ -150,7 +150,15 @@ or
...
 
             let field_ty = field.field_type()?;
-            let field_ty = if field_ty.is_builtin(type_system)? {
+            let field_ty = if let TypeEnum::EntityId(entity_name) = field_ty {
+                if !type_names.contains(entity_name) {
+                    bail!(
+                        "field `{}` of entity `{name}` is of type `Id<{entity_name}>`, but entity `{entity_name}` is undefined",
+                        field.name
+                    );
+                }
```
honzasp, line 159: This validation seems to be quite ad-hoc. For example, we don't seem to check that `E` is a valid entity in `Id<E>[]`, if I understand the code correctly?

honzasp (approved):

PR #1988, opened 2023-01-19
Reflection and validation for routes.

BearLemma:
This PR brings reflection support for requests for our routing endpoints.
The following is now possible:
```
        type Query = {
            count: number;
            name: string;
            human: boolean;
        };
        export default new RouteMap()
            .post('/foo', async (req: ChiselRequest<Query>) => {
                const q = req.typedQuery();
                const count: number = q.count;
                const name: string = q.name;
                const human: boolean = q.human;
                return [count, name, human];
            });
```
The type validity of Query parameters is checked at run-time and sender gets an error if he provides (or doesn't provide required) parameter.

Same goes for Json Body
```
        import { RouteMap, ChiselRequest } from '@chiselstrike/api';
        type Company = {
            name: string
        };
        type Body = {
            name: string,
            age: number,
            human: boolean,
            birthDate: Date,
            employer: Company,
            formerEmployers: Company[],
            school: {
                founder: string
            },
        };
        export default new RouteMap()
            .post('/foo', async (req: ChiselRequest<{}, Body>) => {
                const b = await req.typedJson();
                const name: string = b.name;
                const age: number = b.age;
                const human: boolean = b.human;
                const birthDate: Date = b.birthDate;
                const employer: Company = b.employer;
                const formerEmployers: Company[] = b.formerEmployers;
                const school: {founder: string} = b.school;
                return {name, age, human, birthDate, employer, formerEmployers, school};
            });
        "#,
```

Some notes and constraints:

1. It's only available in Node mode. Doing it for Deno mode is challenging because the build system of chiselstrike is quite horrible.
2. Works only with arrow functions and supported types are a bit limited (No recursive types, I haven't tested classes)
3. If user doesn't provide the type parameters, the code behaves the same as it did before
4. Should be +- fully backwards compatible (`ChiselRequest` has well behaved type defaults)
5. The reflection is now available to be used for client generation
6. Extending it to handle return types should be very easy.
7. Support added only for `.post` as it's experimental.


penberg:
Looks good. We can skip Deno mode support. Any idea why Kafka test is failing, though?

api/src/request.ts
```diff
@@ -1,6 +1,22 @@
 // SPDX-FileCopyrightText: © 2022 ChiselStrike <info@chiselstrike.com>
 
 import type { AuthUser } from "./datastore.ts";
+import {
+    ChiselError,
+    HTTP_STATUS,
+    JSONValue,
+    ReflectionType,
```
honzasp, line 8: I think that types should be imported using `import type`, so that we can transpile the .ts files into .js using swc. (Isn't there a linter that warns about this?)

tsc_reflection/src/reflection.ts
```diff
@@ -0,0 +1,59 @@
...
+        symbol.getName() === "Date" &&
+        type.getProperty("getTime") !== undefined
+    ) {
+        return { name: "date" };
+    } else {
+        const fields: Record<string, ReflectionType> = {};
+        for (const property of type.getProperties()) {
+            const name = property.getName();
+            const propertyType = tc.getTypeAtLocation(property.getValueDeclarationOrThrow());
+            fields[name] = getTypeReflection(tc, propertyType);
```
honzasp, line 48: What happens for recursive types like `type List = {head: number, tail: List}`? Will we fall into an infinite loop?

tsc_reflection/src/main.ts
```diff
@@ -0,0 +1,133 @@
...
+            modifyCallExpression(tc, node as tsm.CallExpression);
+        }
+    });
+    await srcFile.save();
+}
+
+function modifyCallExpression(tc: tsm.TypeChecker, callExpr: tsm.CallExpression) {
+    const propertyAccess = callExpr.getExpression().asKind(tsm.SyntaxKind.PropertyAccessExpression);
+    if (propertyAccess !== undefined) {
+        if (propertyAccess.getName() === "post") {
```
honzasp, line 47: As we talked previously, we should check that the receiver is a RouteMap.

tsc_reflection/src/main.ts
```diff
@@ -0,0 +1,133 @@
...
+    project.addSourceFilesAtPaths([path.join(routesDir, "/**/*{.d.ts,.ts}")]);
+    project.resolveSourceFileDependencies();
+
+    const diagnostics = project.getPreEmitDiagnostics();
+    for (const diag of diagnostics) {
+        console.log(diag.getMessageText());
+    }
+
+    const tc = project.getTypeChecker();
+    tc.getApparentType;
```
honzasp, line 25: This seems to be a leftover of some refactoring?

tsc_reflection/src/main.ts
```diff
@@ -0,0 +1,133 @@
...
+): HandlerReflection | undefined {
+    const args = callExpr.getArguments();
+    assertEquals(args.length, 2, "Unexpected number of call arguments");
+    const handlerArg = args[1];
+
+    const arrowHandler = handlerArg.asKind(tsm.SyntaxKind.ArrowFunction);
+    if (arrowHandler !== undefined) {
+        const callSignature = arrowHandler.getSignature();
+        const params = callSignature.getParameters();
+        assertEquals(params.length, 1, "Endpoint handler must have only one argument");
```
honzasp, line 77: Very simple handlers may have zero parameters if they don't need to access the request at all.

tsc_reflection/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+#[derive(Debug)]
+struct ModuleLoader;
+
+impl deno_runtime::deno_core::ModuleLoader for ModuleLoader {
+    fn resolve(&self, specifier: &str, _referrer: &str, _is_main: bool) -> Result<Url> {
+        if specifier == "chisel://generate_reflection.js" {
+            Ok(Url::from_str(specifier)?)
+        } else {
+            panic!("Only generate_reflection.js specifier supported (everything should be included in that bundle)");
+        }
```
honzasp, line 25: It might be a bit more cleaner to just call `deno_core::resolve_import()` here, and check for `generate_reflection.js` in the `load()` method.

tsc_reflection/build.rs
```diff
@@ -0,0 +1,26 @@
...
+    let bundle_path = out_dir.join("generate_reflection.js");
+
+    let x = std::process::Command::new("deno")
+        .args(vec![
+            "bundle",
+            "src/main.ts",
+            &bundle_path.to_string_lossy(),
+        ])
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
```
honzasp, line 17: Maybe it would be better to just let `deno bundle` inherit stdout/stderr from the current process? In this way, stdout and stderr from `deno bundle` would be correctly interleaved.

cli/src/cmd/apply/common.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+    Ok(file_map)
+}
+
+pub(crate) fn create_tmp_topic_files(
+    mut file_map: FileTopicMap,
+    gen_dir: &Path,
+) -> Result<FileTopicMap> {
+    let cwd = env::current_dir()?;
+    for route in file_map.topics.iter_mut() {
```
honzasp, line 25: s/`route`/`topic`

honzasp (commented):
Nice demo! It makes the build system slightly more convoluted, but it probably doesn't matter anymore at this point.

tsc_reflection/src/main.ts
honzasp: We do, that's how JavaScript (and, by extension, TypeScript) works: if you pass extra arguments to a function, they will be simply ignored, so a function `() => ResponseLike` is a subtype of `(req: ChiselRequest) => ResponseLike`.

tsc_reflection/build.rs
```diff
@@ -0,0 +1,24 @@
...
+    let out_dir = PathBuf::from(env::var_os("OUT_DIR").unwrap());
+    let bundle_path = out_dir.join("generate_reflection.js");
+
+    let x = std::process::Command::new("deno")
+        .args(vec![
+            "bundle",
+            "src/main.ts",
+            &bundle_path.to_string_lossy(),
+        ])
+        .output()
```
honzasp, line 16: Instead of capturing stdout and stderr, it might be better to let the child process inherit stdout and stderr from the build script. It might be less confusing and will keep stdout and stderr properly interleaved.

To do this, you can just replace `output()` with `status()`:
https://doc.rust-lang.org/std/process/struct.Command.html#method.status

honzasp (approved):

